PR #27, opened 2016-08-17
Fixing buildDir and allowing people to add their own version forcing

ethankhall:
With the buildDir change pygradle will respect what users set to be their buildDir for the project. Before it would not honor it for everything.

With the pinned dependency update users can how override our default values or add their own so they can pick build libraries that are used.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
@@ -62,7 +62,7 @@ class PythonExtension {
     public ConsoleOutput consoleOutput = ConsoleOutput.ASCII
 
     public PythonExtension(Project project) {
-        this.details = new PythonDetails(project, new File(project.buildDir, "venv"))
+        this.details = new PythonDetails(project)
         docsDir = project.file("${project.projectDir}/docs").path
         testDir = project.file("${project.projectDir}/test").path
         srcDir = project.file("${project.projectDir}/src").path
@@ -70,7 +70,7 @@ class PythonExtension {
         pinnedFile = project.file("pinned.txt")
 
         pythonEnvironment = [
-                'PATH': project.file("${details.virtualEnv.absolutePath}/bin").path + ':' + System.getenv('PATH'),]
+                'PATH': "${-> details.virtualEnv.absolutePath}/bin" + ':' + System.getenv('PATH'),]
 
         pythonEnvironmentDistgradle = ['PYGRADLE_PROJECT_NAME'   : project.name,
                                        'PYGRADLE_PROJECT_VERSION': "${ -> project.version }",]
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
@@ -56,6 +56,7 @@ class PythonExtension {
     /** The name of the pinned requirements file. */
     public File pinnedFile
 
+    /** A way to define forced versions of libraries */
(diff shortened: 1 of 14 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/FileSystemUtils.java
```diff
@@ -33,18 +31,16 @@ private FileSystemUtils() {
...
-    public static void makeSymLink(Project project, File target, File destination) throws IOException {
+    public static void makeSymLink(File target, File destination) throws IOException {
         /*
          * Check if the file exists because the link checking logic in Gradle differs
          * between Linux and OS X machines.
          */
-        if (!project.file(destination).exists()) {
-            Files.createSymbolicLink(destination.toPath(), target.toPath());
-        }
+        Files.deleteIfExists(destination.toPath());
```
zvezdan: Wouldn't this remove users' files without their knowledge?
Seems destructive compared to previous implementation.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/PythonPlugin.groovy
```diff
@@ -189,7 +164,11 @@ class PythonPlugin implements Plugin<Project> {
 
             doLast {
                 def activateLinkSource = VirtualEnvExecutableHelper.getExecutable(settings, "bin/activate")
-                FileSystemUtils.makeSymLink(project, activateLinkSource, settings.getDetails().activateLink)
+                def activateLink = settings.getDetails().activateLink
+                if (activateLink.exists()) {
+                    activateLink.delete()
```
zvezdan: What if user uses the same name by mistake?
We should not delete files like this and should fail instead gracefully so that user learns to avoid using the same name we use for this symbolic link.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
```diff
@@ -56,21 +56,39 @@ class PythonExtension {
...
+        'pex'           : ['group': 'pypi', 'name': 'pex', 'version': '1.1.4'],
+        'pip'           : ['group': 'pypi', 'name': 'pip', 'version': '7.1.2'],
+        'pytest'        : ['group': 'pypi', 'name': 'pytest', 'version': '2.9.1'],
+        'pytest-cov'    : ['group': 'pypi', 'name': 'pytest-cov', 'version': '2.2.1'],
+        'pytest-xdist'  : ['group': 'pypi', 'name': 'pytest-xdist', 'version': '1.14'],
+        'setuptools'    : ['group': 'pypi', 'name': 'setuptools', 'version': '19.1.1'],
+        'setuptools-git': ['group': 'pypi', 'name': 'setuptools-git', 'version': '1.1'],
+        'six'           : ['group': 'pypi', 'name': 'six', 'version': '1.10.0'],
+        'Sphinx'        : ['group': 'pypi', 'name': 'Sphinx', 'version': '1.4.1'],
+        'unittest2'     : ['group': 'pypi', 'name': 'unittest2', 'version': '1.1.0.1'],
```
zvezdan: This is an internal version (1.1.0.1)

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
```diff
@@ -56,21 +56,39 @@ class PythonExtension {
     /** The name of the pinned requirements file. */
     public File pinnedFile
 
+    public Map<String, Map<String, String>> forcedVersions = [
```
zvezdan: Can we add a doc comment for the public field here?

PR #31, opened 2016-08-24
Cannot set `details` block in `build.gradle`

worr:
When setting the `details` block in your `build.gradle`, gradle yields the following when trying to build your project.

```
FAILURE: Build failed with an exception.

* Where:
Build file '/tmp/junit3053109164895062324/build.gradle' line: 7

* What went wrong:
A problem occurred evaluating root project 'junit3053109164895062324'.
> Could not find method details() for arguments [build_4vooep3cvvjfq4yfrz6qrhcgl$_run_closure1$_closure3@68f7c8eb] on root project 'junit3053109164895062324'.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED

Total time: 0.152 secs

	at org.gradle.testkit.runner.internal.DefaultGradleRunner$1.execute(DefaultGradleRunner.java:222)
	at org.gradle.testkit.runner.internal.DefaultGradleRunner$1.execute(DefaultGradleRunner.java:219)
	at org.gradle.testkit.runner.internal.DefaultGradleRunner.run(DefaultGradleRunner.java:282)
	at org.gradle.testkit.runner.internal.DefaultGradleRunner.build(DefaultGradleRunner.java:219)
	at com.linkedin.gradle.python.plugin.PythonPluginIntegrationTestv3.can build library(PythonPluginIntegrationTestv3.groovy:67)
```

I've attached a PR with an integration test to demonstrate the failure.

### DIFF
--- a/pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PythonPluginIntegrationTest.groovy
+++ b/pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PythonPluginIntegrationTest.groovy
@@ -96,4 +96,43 @@ class PythonPluginIntegrationTest extends Specification {
         result.task(':check').outcome == TaskOutcome.SUCCESS
         result.task(':build').outcome == TaskOutcome.SUCCESS
     }
+
+    def "can build v3 library"() {
+        given:
+        testProjectDir.buildFile << """
+        |plugins {
+        |    id 'com.linkedin.python'
+        |}
+        |
+        |python {
+        | details {
+        |   pythonVersion = 3.5
+        | }
+        |}
+        |
+        |${PyGradleTestBuilder.createRepoClosure()}
+        """.stripMargin().stripIndent()
+
+        when:
+        def result = GradleRunner.create()
+                .withProjectDir(testProjectDir.root)
+                .withArguments('build')
+                .withPluginClasspath()
+                .withDebug(true)
+                .build()
+        println result.output
+
+        then:
+
+        result.output.contains("BUILD SUCCESS")
+        result.output.contains('test/test_a.py .')
+        result.task(':flake8').outcome == TaskOutcome.SUCCESS
+        result.task(':installPythonRequirements').outcome == TaskOutcome.SUCCESS
+        result.task(':installTestRequirements').outcome == TaskOutcome.SUCCESS
+        result.task(':createVirtualEnvironment').outcome == TaskOutcome.SUCCESS
+        result.task(':installProject').outcome == TaskOutcome.SUCCESS
+        result.task(':pytest').outcome == TaskOutcome.SUCCESS
+        result.task(':check').outcome == TaskOutcome.SUCCESS
+        result.task(':build').outcome == TaskOutcome.SUCCESS
+    }
 }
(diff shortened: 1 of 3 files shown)


worr:
Added a rudimentary fix

pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PythonPluginIntegrationTest.groovy
```diff
@@ -96,4 +96,43 @@ class PythonPluginIntegrationTest extends Specification {
...
+    def "can build v3 library"() {
+        given:
+        testProjectDir.buildFile << """
+        |plugins {
+        |    id 'com.linkedin.python'
+        |}
+        |
+        |python {
+        | details {
+        |   pythonVersion = 3.5
```
zvezdan: This is not a correct setting.
It should be '3.5' -- string, not a float.

pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PythonPluginIntegrationTest.groovy
```diff
@@ -96,4 +96,43 @@ class PythonPluginIntegrationTest extends Specification {
         result.task(':check').outcome == TaskOutcome.SUCCESS
         result.task(':build').outcome == TaskOutcome.SUCCESS
     }
+
+    def "can build v3 library"() {
+        given:
+        testProjectDir.buildFile << """
+        |plugins {
```
zvezdan: Space between the margin `|` and the text is needed.
The indent should stay 4 spaces throughout, even for sample code.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
```diff
@@ -81,6 +84,7 @@ class PythonExtension {
 
     public PythonExtension(Project project) {
         this.details = new PythonDetails(project)
+        this.project = project
```
zvezdan: This class does not use the `project` outside the constructor.
It simply passes it to its parts, including the details.
We probably don't need this as a private variable.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
```diff
@@ -129,6 +133,10 @@ class PythonExtension {
         return details
     }
 
+    public void details(Closure cl) {
```
zvezdan: I suspect that all we need here is a setter method `setDetails` instead of this public member.
We already have a private class field for details, but it is currently immutable (final) -- hence no setter.
Let's see what others think about this approach.

ethankhall: This is the correct pattern (ref: https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#dependencies(groovy.lang.Closure) ). You should call it the name you want in the DSL and it accepts a Closure.

pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PythonPluginIntegrationTest.groovy
ethankhall: Groovy is converting the float to a String automatically.

zvezdan: I know, but the integration tests show the usage pattern too.
And we would prefer to keep that pattern match the intended usage.

ethankhall:
Moved this to #36. Closing this PR.

PR #33, opened 2016-08-25
Changing url to https://linkedin.jfrog.io/linkedin/pypi-external/

ethankhall:
Bintray is going to delete the repo I set up because of a trial. So
instead of just moving it to a different name quickly lets make it a
good solution that will scale.


sholsapp:
I think changes related to upgrading to Gradle 3.0 should be separate from changes related to our repository's URL.

ethankhall:
Splitting it up into multiple PRs

Issue #26, opened 2016-08-17
Convert codebase to java8


ethankhall:
In an effort to make sure that the code is stable and doesn't get small bugs, lets try to move most (not all) of the code into Java. For the code that already is Java lets fix it so it matches the Java8 style coming from Java6.

Shad0w1nk:
I could help with this a bit while I learn the plugin. I currently have two questions regarding this work:
1. Do we care about backward compatibility regarding the ABI?
2. Would you consider this work as incremental PRs (like one or more class at a time)?

ethankhall:
By ABI to you mean the JNI?

Yep! If you want to make small changes open a review and we'll merge them as they come in, no reason to hold onto incremental updates :-)

Shad0w1nk:
I read this in the Gradle 3.0 release note [1] (see section "Groovy to Java conversions"). There is ways to work around if compatibility is required. We could use the same workaround as Gradle use and once the conversion is finished, we can remove the fix in the next major release. It real depends on how you feel about honoring the compatibility between version (if it's a strict semantic versioning [2]).

[1] https://docs.gradle.org/3.0/release-notes
[2] http://semver.org/

ethankhall:
I don't think it will be a big deal. I've been converting classes for a while to Java and haven't ran into the issue. We are also on 0.3 so this is the time to do these changes.

Shad0w1nk:
Sounds like a good plan.

oesolutions:
My 2 cents on forcing Java 8 on plugin users; Please consider supporting the same Java versions as Gradle, without requiring workarounds.
This means, if you are supporting Gradle 2.x, then support Java 6 and up. If you only intend on supporting Gradle 3+, then support Java 7 and up.

Personally I think Netflix made a mistake in dropping support for running their plugins in Java 6. IMO, it is much more community-friendly if you support the same Java versions as Gradle does.
For us, Netflix's decision has caused nothing but pain, blocking us from upgrading their plugins and, even worse, from upgrading Gradle due to some of their plugin's usage of Gradle internals.
Using workarounds to support Java 6 may seem easy, but we've found that quite challenging at scale, despite our efforts in using "enterprise Gradle" techniques.

Shad0w1nk:
[...]

PR #34, opened 2016-08-25
First steps at moving to java 7 and Gradle 3

ethankhall:
Also upgraded the build to use Gradle 3.0

And to take a first pass at #26 I set the Java language level to 1.7 and
fixed the errors that came with compiling at 1.8. This is an effort to
make sure that we are running on what Gradle is using with 3.0.


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/internal/pex/ThinPexGenerator.java
```diff
@@ -44,7 +44,7 @@
     public ThinPexGenerator(Project project, String entryPointTemplate, Map<String, String> extraProperties) {
         this.project = project;
         this.entryPointTemplate = entryPointTemplate;
-        this.extraProperties = extraProperties == null ? new HashMap<>() : extraProperties;
+        this.extraProperties = extraProperties == null ? new HashMap<String, String>() : extraProperties;
```
zvezdan: Java 7 introduced the diamond in the cases like this `HashMap<>` and will infer the types correctly.
Do we really need to repeat the types that are already declared in the field itself?

ethankhall: Without the types I get the following error compiling
`error: incompatible types: Map<? extends Object,? extends Object> cannot be converted to Map<String,String>`. 

This has something to do with the turnary statement, if i convert it to 
```
if(null == extraProperties) {
    this.extraProperties = new HashMap<>();
} else {
    this.extraProperties = extraProperties;
}
```
that works, do you think the expanded format is more usable/readable/perferable?

zvezdan: No. Leave it as is then. It seems that Java 8 improved even the diamond inference. That's why I preferred that we use Java 8. Oh, well ... :-)

PR #36, opened 2016-08-25
Fixing dsl for python.details {}

ethankhall:
Moving from using project to setting the deletage on the closure

This should also make sure that if people have pygradle loaded their
groovy IDE should be able to let you know what the options are for the
closure.

This removes an integ test and adds a unit test. The integ test sets a
value to make sure the parse works, but doesn't require another test
(which adds a bunch of time to builds)

This fixes the issues in #31 and supersedes it.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
@@ -31,9 +31,6 @@ class PythonExtension {
     /** The environment to use for all Python commands. */
     public Map<String, Object> pythonEnvironment
 
-    /** This gradle project */
-    private final Project project
-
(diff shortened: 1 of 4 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
```diff
@@ -129,6 +129,12 @@ class PythonExtension {
         return details
     }
 
+    public void details(@DelegatesTo(PythonDetails) Closure cl) {
```
sholsapp: Please add documentation.

ethankhall:
This is failing because #35 needs to be done first. It adds the travis config file to make it build.

zvezdan:
Go ahead and rebase to get the checks passing.

PR #39, opened 2016-09-01
Fixing spelling and pip install wierdness

ethankhall:
Was looking at a build log and realized there was a typo:

```
Prepairing wheel futures-3.0.1 ...................................... [STARTING]
Preparing wheel futures-3.0.1 (0:01 s) .............................. [FINISHED]
Prepairing wheel gunicorn-19.3.0 .................................... [STARTING]
Preparing wheel gunicorn-19.3.0 (0:01 s) ............................ [FINISHED]
Prepairing wheel ipaddress-1.0.16 ................................... [STARTING]
```

This change also fixes where when the project installs other projects built by pygradle using the PYGRADLE_NAME and PYGRADLE_VERSION settings they will override what the project wants to use and causes weird install issues.

### DIFF
--- a/gradle/publishing.gradle
+++ b/gradle/publishing.gradle
@@ -1,4 +1,5 @@
 apply plugin: 'maven-publish'
+apply plugin: 'ivy-publish'
 apply plugin: 'com.jfrog.artifactory'
 
 task sourceJar(type: Jar) {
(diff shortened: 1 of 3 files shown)


gradle/publishing.gradle
```diff
@@ -42,6 +43,19 @@ publishing {
                 }
             }
         }
+        ivyJava(IvyPublication) {
```
sholsapp: Is this change related to the one your commit message talks about?

ethankhall: Yep.

PR #44, opened 2016-09-08
Porting features from internal divergence into open source

ethankhall:
This change makes it where things operate off the PythonDetails object
vs the PythonExtension object. This will allow the creation of multiple
different venvs where they contain different versions of python.

This also makes it where the PipInstallTasks have a way to filter
dependencies, this will allow us to use extra-requires filters to
add/remove dependencies.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PyTestTask.groovy
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PyTestTask.groovy
@@ -46,6 +46,8 @@ class PyTestTask extends AbstractPythonTestSourceDefaultTask {
     void processResults(ExecResult execResult) {
         if (execResult.exitValue == NO_TESTS_COLLECTED_ERRNO) {
             logger.warn("***** WARNING: You did not write any tests! *****")
+        } else {
+            execResult.assertNormalExitValue()
         }
     }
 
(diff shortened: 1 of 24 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PyTestTask.groovy
```diff
@@ -47,8 +46,6 @@ class PyTestTask extends AbstractPythonTestSourceDefaultTask {
     void processResults(ExecResult execResult) {
         if (execResult.exitValue == NO_TESTS_COLLECTED_ERRNO) {
             logger.warn("***** WARNING: You did not write any tests! *****")
-        } else {
```
zvezdan: Do we still need `else` and call to `super.processResults(execResult)` or some other form of action here when `exitValue` is non-zero and not `NO_TESTS_COLLECTED_ERRNO`? We need to raise an exception at least and `assertNormalExitValue` was ensuring that. I think it's still necessary.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/PexFileUtil.groovy
```diff
@@ -78,6 +78,11 @@ class PexFileUtil {
                     | This typically happens because your virtual environment contains a cached copy of ${packageName}
                     | that no other package depends on any more.
                     | Usually, this is the result of updating a package that used to depend on ${packageName}.
+                    |
+                    | Another possible reason is that you started using a newer version of Python
+                    | without checking if all the libraries you use are compatible.
+                    | If you can find the missing package in the output above with the message
+                    | [EXCLUDED], then it works only with lower versions of Python.
```
ethankhall: Do we need the cast?

PR #53, opened 2016-09-22
Add special cases.

zvezdan:
We have two places in the code where these are handled.
We added them in one place only and forgot to add it to the second.
Later, we need to remove this duplication.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/internal/pex/PipFreezeAction.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/internal/pex/PipFreezeAction.java
@@ -50,7 +50,14 @@ class PipFreezeAction {
         developmentDependencies.addAll(configurationToSet(project, PythonPlugin.CONFIGURATION_TEST));
 
         // Special cases, such as sphinx-rtd-theme with weird metadata
-        developmentDependencies.addAll(Arrays.asList("sphinx-rtd-theme", "sphinx_rtd_theme"));
+        developmentDependencies.addAll(Arrays.asList(
+            "sphinx-rtd-theme",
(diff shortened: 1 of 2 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/internal/pex/PipFreezeAction.java
```diff
@@ -50,7 +50,14 @@ public PipFreezeAction(Project project) {
         developmentDependencies.addAll(configurationToSet(project, PythonPlugin.CONFIGURATION_TEST));
 
         // Special cases, such as sphinx-rtd-theme with weird metadata
-        developmentDependencies.addAll(Arrays.asList("sphinx-rtd-theme", "sphinx_rtd_theme"));
+        developmentDependencies.addAll(Arrays.asList(
```
sholsapp: What do you think we should do to handle this gracefully into the future? It'd be painful to have to maintain a list like this indefinitely, although at the moment we don't have any other options. Just curious if you have ideas.

zvezdan: I have an algorithm ready for testing, but we need this fix to merge Ethan's branch internally.
I'll add a better fix soon.

sholsapp: Got it. Thanks.

zvezdan: I just implemented the fix. I'm going to discard this PR.

PR #61, opened 2016-09-30
Fix incorrect dependency to pypi:requests

DALDEI:
Fixed dependency from pypi:requests 2.5.1 to 2.9.1


ethankhall:
Thanks @DALDEI for the catch.

PR #64, opened 2016-10-04
Raise a specific exception so that it can be caught.

zvezdan:
Added a custom MissingInterpreterException class.


sholsapp:
LGTM but will wait for checks to pass.

PR #65, opened 2016-10-13
Add access to individual PythonVersion parts.

sholsapp:
Exposing each part of the PythonVersion (i.e., major, minor, patch)
allows us to write code that can use this class to format other strings
that make use of a Python version.

E.g., one use case is wheel naming, which uses a Python version, but
doesn't format the version like we do currently.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonVersion.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonVersion.java
@@ -25,6 +25,11 @@ public class PythonVersion implements Serializable {
         this.version = version;
     }
 
+    private String getVersionPart(int position) {
+        String[] parts = version.split("\\.");
+        return parts[position];
+    }
+
     /**
      * @return The exact version of Python this project uses, such as '2.7.11'.
      */
@@ -36,14 +41,27 @@ public class PythonVersion implements Serializable {
      * @return The short version of Python this project uses, such as '2.7'.
      */
     public String getPythonMajorMinor() {
-        String[] split = version.split("\\.");
-        return String.format("%s.%s", split[0], split[1]);
+        return String.format("%s.%s", getVersionPart(0), getVersionPart(1));
     }
 
     /**
      * @return The major version of Python this project uses, such as '2'.
      */
     public String getPythonMajor() {
-        return version.split("\\.")[0];
+        return getVersionPart(0);
+    }
+
+    /**
+     * @return The minor version of Python this project uses, such as '7'.
+     */
+    public String getPythonMinor() {
+        return getVersionPart(1);
+    }
+
+    /**
+     * @return The patch version of Python this project uses, such as '11'.
+     */
+    public String getPythonPatch() {
+        return getVersionPart(2);
     }
 }


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonVersion.java
```diff
@@ -19,31 +19,53 @@
...
+    private final String minor;
+    private final String patch;
 
     public PythonVersion(String version) {
-        this.version = version;
+        String[] split = version.split("\\.");
+        this.major = split[0];
+        this.minor = split[1];
+        if (split.length < 3) {
+            this.patch = "0";
```
sholsapp: @zvezdan let me know if this is a bad assumption. CPython has been versioned with major/minor for a long time, so we might want to set this to `null` instead of `0`. Not sure.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonVersion.java
```diff
@@ -25,25 +25,43 @@ public PythonVersion(String version) {
...
+        String[] parts = this.version.split("\\.");
+        return parts[position];
+    }
+
     /**
      * @return The exact version of Python this project uses, such as '2.7.11'.
      */
     public String getPythonVersion() {
-        return version;
+        return this.version;
```
zvezdan: No need for `this` here. There's no ambiguity with the name as it happens in the setter or constructor where the parameter is also named `version`. Here, in the getter, it can only be our internal, private field `version`. Let's leave this method unchanged.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonVersion.java
```diff
@@ -25,25 +25,43 @@ public PythonVersion(String version) {
         this.version = version;
     }
 
+    private String getVersionPart(int position) {
+        String[] parts = this.version.split("\\.");
```
zvezdan: No need for `this.` here. There are no other arguments or variables named `version`.
These two lines can be reduced to `return version.split("\\.")[position];`
Could you move this private method down below the public API?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonVersion.java
```diff
@@ -25,25 +25,43 @@ public PythonVersion(String version) {
...
+        return this.version;
     }
 
     /**
      * @return The short version of Python this project uses, such as '2.7'.
      */
     public String getPythonMajorMinor() {
-        String[] split = version.split("\\.");
-        return String.format("%s.%s", split[0], split[1]);
+        return String.format("%s.%s", getVersionPart(0), getVersionPart(1));
```
zvezdan: Hmm. Two method calls with two splits instead of one split. I'm inclined to leave this method unchanged. It's working just fine. Although, I can see why you would want the consistency and isolation of this handling in the `getVersionPart` only.

sholsapp:
I think this one is ready to go @zvezdan

PR #67, opened 2016-10-25
Enable junit test reporting by default

dnozay:
Providing a jUnit xml report by default is useful for continuous integration systems.

### DIFF
--- a/pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PyGradleTestBuilder.groovy
+++ b/pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PyGradleTestBuilder.groovy
@@ -24,7 +24,7 @@ class PyGradleTestBuilder {
             | max-line-length = 160
             |
             | [pytest]
-            | addopts = --ignore build/ --ignore dist/
+            | addopts = --ignore build/ --ignore dist/ --junitxml TEST-pytest.xml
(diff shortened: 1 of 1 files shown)


ethankhall:
You changes how the local integration tests work, is that what you were intending? I think you probably want to make it so that by default the PyTestTask adds that to the argument list, or better make it where people can specify one of the N options that are avaliable.

pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PyGradleTestBuilder.groovy
```diff
@@ -24,7 +24,7 @@ class PyGradleTestBuilder {
             | max-line-length = 160
             |
             | [pytest]
-            | addopts = --ignore build/ --ignore dist/
+            | addopts = --ignore build/ --ignore dist/ --junitxml TEST-pytest.xml
```
zvezdan: I don't think this will help CI systems in any way. This is the test for {py}gradle itself.
Let's talk offline about your needs and how to best solve them.
Could you please delete this branch from the project?
We prefer that the development for pull requests be done on the forks of the project, rather than directly on the master repo.

dnozay: I need to close the pull request to delete the branch.

PR #71, opened 2016-11-05
Use `tabby-mctabface` to complete some parameters.

anonfunc:
- `tabby-mctabface list` must return a whitespace separated list of
  parameter names with custom completion.
- A parameter (example) has a corresponding function `_example` which is
	available in the user's shell.

If `tabby-mctabface` is missing when the completions are generated, it is
not used in the generated completion files.


zvezdan:
Can users of open-source {py}gradle ever have tabby-mctabface available?
If not, we should put this in the internal plugin as an override.
We've done that for some other features already.

kopy07:
This has been open for a year without any comments. Closing.

PR #81, opened 2016-11-16
packageSdist should on incremental builds if source changes. Fixes #80

### LINKED ISSUE #80: sdist does not get rebuilt when source changes
The sdist task(SourceDistTask.java) is not properly configured with inputs/outputs to detect changes in source, for example. Deriving from AbstractPythonMainSourceDefaultTask instead of DefaultTask(present case) fixes this problem.

We should also exclude **/*.egg-info/ from source Inputfiles.

schakrava:
- extend from AbstractPythonMainSourceDefaultTask instead of DefaultTask
  for correct Inputs.
- Exclude *.egg-info directories from Inputfiles.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/AbstractPythonMainSourceDefaultTask.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/AbstractPythonMainSourceDefaultTask.java
@@ -68,7 +68,7 @@ abstract public class AbstractPythonMainSourceDefaultTask extends DefaultTask im
     }
 
     public String[] standardExcludes() {
-        return new String[]{"**/*.pyc", "**/*.pyo", "**/__pycache__/"};
+        return new String[]{"**/*.pyc", "**/*.pyo", "**/__pycache__/", "**/*.egg-info/"};
     }
 
     @Internal
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/SourceDistTask.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/SourceDistTask.java
@@ -17,39 +17,19 @@ package com.linkedin.gradle.python.tasks;
 
 import java.io.File;
 
-import com.linkedin.gradle.python.PythonExtension;
-import com.linkedin.gradle.python.tasks.execution.FailureReasonProvider;
-import com.linkedin.gradle.python.tasks.execution.TeeOutputContainer;
-import com.linkedin.gradle.python.util.VirtualEnvExecutableHelper;
-import org.gradle.api.Action;
-import org.gradle.api.DefaultTask;
 import org.gradle.api.Project;
 import org.gradle.api.tasks.OutputFile;
-import org.gradle.api.tasks.TaskAction;
-import org.gradle.process.ExecSpec;
+import org.gradle.process.ExecResult;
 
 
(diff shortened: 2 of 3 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/SourceDistTask.java
```diff
@@ -22,14 +22,14 @@
...
 import org.gradle.api.tasks.TaskAction;
+import org.gradle.process.ExecResult;
 import org.gradle.process.ExecSpec;
 
 
-public class SourceDistTask extends DefaultTask implements FailureReasonProvider {
+public class SourceDistTask extends AbstractPythonMainSourceDefaultTask implements FailureReasonProvider {
 
     private TeeOutputContainer container = new TeeOutputContainer();
     @TaskAction
```
ethankhall: If we're switching to use AbstractPythonMainSourceDefaultTask let's use it all the way.

We can get rid of the `packageSdist()` method and create a public constructor that adds the args like we do in `com.linkedin.gradle.python.tasks.Flake8Task`

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/AbstractPythonMainSourceDefaultTask.java
```diff
@@ -68,7 +68,7 @@ public FileCollection getSourceFiles() {
     }
 
     public String[] standardExcludes() {
-        return new String[]{"**/*.pyc", "**/*.pyo", "**/__pycache__/"};
+        return new String[]{"**/*.pyc", "**/*.pyo", "**/__pycache__/", "**/*.egg-info/"};
```
ethankhall: Awesome Thanks!

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/SourceDistTask.groovy
```diff
@@ -0,0 +1,40 @@
...
+import org.gradle.api.tasks.OutputFile
+import org.gradle.process.ExecResult
+
+@CompileStatic
+public class SourceDistTask extends AbstractPythonMainSourceDefaultTask {
+
+    private final String distDir = "${project.buildDir}/distributions"
+
+    public void preExecution() {
+    args("setup.py",
```
ethankhall: Minor thing, merge this onto one line please :-)

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/SourceDistTask.groovy
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+package com.linkedin.gradle.python.tasks
+
+import groovy.transform.CompileStatic
+import org.gradle.api.tasks.OutputFile
+import org.gradle.process.ExecResult
+
+@CompileStatic
+public class SourceDistTask extends AbstractPythonMainSourceDefaultTask {
```
ethankhall: Please leave this as a Java class. This will allow us to move to Java 8 in the future.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/SourceDistTask.groovy
schakrava: no problem. Done.

PR #86, opened 2016-12-01
Replace Java8 syntax with Java7

bivas:
Since Gradle 3.x requires support for Java7 I've updated the plugin to use only Java7 syntax

Related to Issue #26

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
@@ -23,6 +23,7 @@ import com.linkedin.gradle.python.util.ConsoleOutput
 import com.linkedin.gradle.python.util.ExtensionUtils
 import com.linkedin.gradle.python.util.PackageInfo
 import com.linkedin.gradle.python.util.VirtualEnvExecutableHelper
+import groovy.time.TimeCategory
 import org.gradle.api.DefaultTask
 import org.gradle.api.GradleException
 import org.gradle.api.Project
@@ -33,8 +34,6 @@ import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.TaskAction
 import org.gradle.process.ExecResult
 import org.gradle.process.ExecSpec
-
-import java.util.concurrent.TimeUnit
 
 class BuildWheelsTask extends DefaultTask {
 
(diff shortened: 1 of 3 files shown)


bivas:
I've tried running this plugin with environment which has JDK7 and the groovy files caused build failures with `java.util.function.Consumer` not found (as it only exists in JDK8).

Groovy still needs to compile to Java. And the syntax I removed is not Groovy, but Java specific e.g. look at the replacement of `forEach` with `each` (which is true Groovy syntax).

bivas:
@zvezdan If you like, I can use Groovy's `TimeDuration` instead of Java8 Duration.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
```diff
@@ -164,7 +163,7 @@ class BuildWheelsTask extends DefaultTask {
                 if (pythonExtension.consoleOutput == ConsoleOutput.RAW) {
                     LOGGER.lifecycle(stream.toString().trim())
                 } else {
-                    String prefix = String.format(messageHead + ' (%d:%02d.%03d s)', duration.toMinutes(), duration.getSeconds() % 60, (int)(duration.getNano() / 1000000))
+                    String prefix = String.format("Install (%d s)", TimeUnit.MILLISECONDS.toSeconds(duration))
```
zvezdan: Why was `messageHead` changed to `"Install"` here?
The message was different in this task compared to the other file below.

zvezdan:
@bivas I'll let @ethankhall decide on the preference between Groovy/Java features used.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
```diff
@@ -164,7 +162,7 @@ class BuildWheelsTask extends DefaultTask {
                 if (pythonExtension.consoleOutput == ConsoleOutput.RAW) {
                     LOGGER.lifecycle(stream.toString().trim())
                 } else {
-                    String prefix = String.format(messageHead + ' (%d:%02d.%03d s)', duration.toMinutes(), duration.getSeconds() % 60, (int)(duration.getNano() / 1000000))
+                    String prefix = String.format(messageHead + " (%d:%02d s)", duration.minutes, duration.seconds % 60)
```
zvezdan: Are we losing milliseconds now?

zvezdan: Are we losing milliseconds now?
Also on the similar line in the other file below.

zvezdan:
@ethankhall Can you also take a look before the merge?

ethankhall:
Sorry I was traveling. Looks good, merging.

PR #90, opened 2016-12-09
Will lazy evaluate vs eager

ethankhall:
This was done because some use-cases (Windows) don't have
/usr/bin/python and thus would never be able to build. So this way the
user can get in early and specify their own python interpreter or update
the path so the plugin can find python. It will still default to
/usr/bin/python if nothing matches

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonDetails.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonDetails.java
@@ -123,10 +123,6 @@ public class PythonDetails implements Serializable {
 
     public void setSystemPythonInterpreter(String path) {
         pythonInterpreter = new File(path);
-        if (!pythonInterpreter.exists()) {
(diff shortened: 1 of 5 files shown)


zvezdan:
We need to consider the change in behavior of `MissingInterpreterException` here. Let's talk offline.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonDetails.java
```diff
@@ -133,6 +131,17 @@ public void setSystemPythonInterpreter(String path) {
     }
 
     public PythonVersion getPythonVersion() {
+        findPythonWhenAbsent();
         return pythonVersion;
     }
+
+    private void findPythonWhenAbsent() {
+        if (pythonInterpreter == null) {
+            File python = ExecutablePathUtils.getExecutable(searchPath, "python");
```
ethankhall: This is intended to be a fallback. If you specify a version to use, or a specific interpreter it will use that (because pythonInterpreter will be non null).

PR #91, opened 2016-12-09
Add scipy/numpy/sklearn example for IRIS

busche:
Even though the initial example project is great for getting started, a more complex example project is needed to gain more insight into how pygradle can be configured and used.
This example project uses the classical text book example for the IRIS dataset classification task in order to train and deploy a SVM model.


ethankhall:
Merged!

PR #95, opened 2016-12-20
Windows compatibility

scphantm:
Initial pull request to attempt to establish a better windows compatibility.  First, this pull fixes an issue with the plugin finding the python.exe file in windows installs.  I need help with the tests to help with the cross platform issues.

The second thing that needs addressed is the venv code.  It has trouble finding and activating the activate executable and continues to believe the venv was created even when you do a clean first.  I will attempt to address that tomorrow but i will likely need help there as my time is very limited on this.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/SystemExecutables.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/SystemExecutables.java
@@ -1,0 +1,36 @@
+package com.linkedin.gradle.python.util;
+
+import org.apache.tools.ant.taskdefs.condition.Os;
+
+/**
+ * Enum that allows for platform independent resolution of executables
+ */
+public enum SystemExecutables {
+    PYTHON("python", "exe"),
+    PYTHON_VERSION_AWARE("python%s", "exe"),
+    ACTIVATE("activate", "bat");
+
+    private String exeName;
+    private String winExtension;
+
+    SystemExecutables(String exeName, String winExtension) {
+        this.exeName = exeName;
+        this.winExtension = winExtension;
+    }
+
+    public String getExecutable() {
+
+        String exeNameReturn = exeName;
+
+        if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+            exeNameReturn = exeName + "." +  winExtension;
+        }
+
+        return exeNameReturn;
+    }
+
+    public String getExecutable(String versionAware) {
+        String exeNameReturn = getExecutable();
+        return String.format(exeNameReturn, versionAware);
+    }
+}
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/SystemExecutables.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/SystemExecutables.java
@@ -27,7 +27,7 @@ public enum SystemExecutables {
 
         if (Os.isFamily(Os.FAMILY_WINDOWS)) {
             exeNameReturn = exeName + "." +  winExtension;
-        } else if (Os.isFamily(Os.FAMILY_UNIX)) {
+        } else if ((Os.isFamily(Os.FAMILY_UNIX)) || (Os.isFamily(Os.FAMILY_MAC))) {
             if (nixExtension != null) {
                 exeNameReturn = exeName + "." + nixExtension;
             } else {
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/SystemExecutables.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/SystemExecutables.java
@@ -6,16 +6,19 @@ import org.apache.tools.ant.taskdefs.condition.Os;
  * Enum that allows for platform independent resolution of executables
  */
 public enum SystemExecutables {
-    PYTHON("python", "exe"),
-    PYTHON_VERSION_AWARE("python%s", "exe"),
-    ACTIVATE("activate", "bat");
+    PYTHON("python", "exe", null),
+    PYTHON_VERSION_AWARE("python%s", "exe", null),
+    ACTIVATE("activate", "bat", null),
+    TEST_FILE("example", "bat", "sh");
 
     private String exeName;
     private String winExtension;
+    private String nixExtension;
 
-    SystemExecutables(String exeName, String winExtension) {
+    SystemExecutables(String exeName, String winExtension, String nixExtension) {
         this.exeName = exeName;
         this.winExtension = winExtension;
+        this.nixExtension = nixExtension;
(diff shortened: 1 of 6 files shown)


scphantm:
I was thinking after I built this, i suspect there are windows executable errors all thru this project.  I was thinking about moving this check into an enumeration.  that way you can pass the enum and the all the is windows checks are centralized in one place.  I will take a deeper look into that this afternoon and see where i get with it.

Also im glad the CI failed because i was having problems with the VENV stuff too.  Hopefully this gives me a lead as to whats wrong

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/SystemExecutables.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.tools.ant.taskdefs.condition.Os;
+
+/**
+ * Enum that allows for platform independent resolution of executables
+ */
+public enum SystemExecutables {
+    PYTHON("python", "exe", null),
+    PYTHON_VERSION_AWARE("python%s", "exe", null),
+    ACTIVATE("activate", "bat", null),
+    TEST_FILE("example", "bat", "sh");
```
ethankhall: Don't include test code.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/SystemExecutables.java
```diff
@@ -0,0 +1,45 @@
...
+        this.exeName = exeName;
+        this.winExtension = winExtension;
+        this.nixExtension = nixExtension;
+    }
+
+    public String getExecutable() {
+
+        String exeNameReturn = exeName;
+
+        if (Os.isFamily(Os.FAMILY_WINDOWS)) {
```
ethankhall: How do you plan to test this if statement?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/SystemExecutables.java
```diff
@@ -0,0 +1,45 @@
+package com.linkedin.gradle.python.util;
+
+import org.apache.tools.ant.taskdefs.condition.Os;
+
+/**
+ * Enum that allows for platform independent resolution of executables
+ */
+public enum SystemExecutables {
```
ethankhall: I like the abstraction, but I don't know if this is the right abstraction. I'm wondering if a `CrossPlatformExecutable` interface and we would have a PythonExecutable and ActivateExecutable that would implement it, then the python one can have a method `withVersion(String version)`.

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/TestPythonExtension.groovy
```diff
@@ -15,13 +15,18 @@
...
+import spock.lang.IgnoreIf
 import spock.lang.Specification
 
 class TestPythonExtension extends Specification {
 
     def project = new ProjectBuilder().build()
 
+    // Have to ignore this if its a windows machine trying due to the fact its
+    // based on shell
+    @IgnoreIf({ Os.isFamily(Os.FAMILY_WINDOWS) })
```
ethankhall: Could we make it work on windows?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/SystemExecutables.java
scphantm: Thats certainly doable.  The only question I haven't answered yet is how many executables are we talking.  5?  Sure, we can do that.  500? Bad design.  I'll keep things as we are now, once we get closer to having this merged in and everything is functional, we can revisit it.  I will most likely take your advice tho, if there's more than a dozen exe's, i'd be shocked.

scphantm:
I really need some direction on how you want to approach this.  My major problem right now are the unit tests in `PythonDetailsTest`.  There's a few ways to handle them.  Break them out into 2 classes, One for linux/mac and one for windows.  They will be independent and the only way you could run them both is if you executed the CI on appveyor as well.

The second way to do it is to try to make the tests OS universal.  It can be done (i think) but it really ramps up the complexity because from what i can see, I would basically have to figure out a way to mock an OS, something i havn't been able to do as of yet.

scphantm:
Another issue to contend with, What about dev machines that don't have 2.6 and 2.7 installed.  The one i am typing this post on doesn't, it only has 3.5 installed.  I need help coming up with a better testing strategy here.  I am in despirate need of this plugin, but this issue is killing me right now.

thanks

ethankhall:
@scphantm, I would make one for Unix-style OS's and one for Windows. I think having simpler tests would be better than complicated ones.

Do you have a nice way to build a Windows VM so I can also do some investigation / look at the Windows story?

scphantm:
unfortunately no, i don't.  To keep it simple I'm going to have to implement a sorta, configuration by convention type test.  Things like, if you are looking for python 2.7, i will look in "Program Files" and search the path for the folder, but if i can't find it there, it assumes its not installed.  

How to handle the executables still escapes me.  In Linux/mac the version is in the executable itself (python2, python3, python2.7, etc).  in windows its in the folder that changes and the executable is always python.exe.  Makes it MUCH more difficult.

Im almost thinking that its not worth even trying to figure it out.  Have the plugin just use the default python it finds in the path and if you want another, specify it in the local.properties or something.  Either way, i don't see a perfect solution there.

ethankhall:
I think picking up the default python is probably the right idea. The use can set the version of python they want to use in their build.gradle files. Then the build will fail if we can't find that version of python in the PATH.

scphantm:
Now that I've gotten some sleep (those 72 hour stretches don't agree with me anymore) i think i came up with something.  Linux, Mac, and Win have a path environment variable.  I can parse that variable, pull a list of all executables with python in it.  Then I can run your version method to determine the version of that executable instead of trying to figure it out via the file name.  Just do that across the board, That way I will probably be able to make the test universal and not OS dependent.  

The drawback is determining the default to use.  What i mean is on linux/mac, you can call `/usr/bin/python` as your default and its typically a symlink to 2.7.  but on windows, if you search the path, you may find 2 or 3 python.exe's and pythonhome is typically not set on windows.  The powershell console is very simple, it uses the first one it finds in the path and could care less what version it is.  If you want to use 3 instead of 2, put 3 in front of 2 in the path.  I can certainly replicate that, then if the build.gradle specifies a version, i can go looking for it.

ethankhall:
@scphantm I took a stab at Windows support today if you want to take a look. The pex's are running and it _looks_ to be working, but if you want to give it a whirl let me know what you think.

Please note, you'll need to append .py to any pex's based on how windows does it's thing.

scphantm:
I will tomorrow.  Thanks

PR #110, opened 2017-02-01
Upgrade PEX to 1.2.2

dsully:


sixninetynine:
tests failing on: 

>  Could not find compatible interpreter

This error is from `pex` itself, here's the block that raises it: https://github.com/pantsbuild/pex/blob/7e8faa2cd32ade42f6a4306828d44765c4f2a655/pex/bin/pex.py#L463-L494 

Since we don't pass a `--python` arg, this line resolves the interpreter path: 

`    interpreter = PythonInterpreter.get()`

Which grabs the interpreter from sys.executable:

```
  @classmethod
  def get(cls):
    return cls.from_binary(sys.executable)
```

Really not sure why this would be failing in such a way...

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/util/PackageInfoTest.groovy
```diff
@@ -62,10 +62,10 @@ class PackageInfoTest extends Specification {
 
     def 'can parse a windows path'() {
         when:
-        def packageInfo = PackageInfo.fromPath('Z:\\pygradle\\build\\ivy-repo\\pypi\\setuptools\\19.1.1\\setuptools-19.1.1')
```
ethankhall: I don't think this needs to be updated with the repo.

dsully: It can't hurt though.

zvezdan:
Did you check if setuptools-19.1.1 actually works with the packages and products we build internally?
I did and I can confirm it doesn't.
Why is this update needed right now?

dsully:
@zvezdan - there's an incompatibility between the new version of CacheControl and pex (which also uses CacheControl).

https://github.com/pantsbuild/pex/pull/357

zvezdan:
@dsully We upgraded pex, setuptools, and more in the #145 and to higher versions than this PR.
Can we close/discard this?

PR #115, opened 2017-02-10
Linkify shields.

sholsapp:
The shields were linking to nothing, therefore couldn't be clicked. This
adds the right service link to each shield.


sholsapp:
I didn't try clicking on any of the shields in the last commit. My bad! 😱

PR #118, opened 2017-02-24
Update appdirs.

zvezdan:
Setuptools picks this from PyPI as the newest version.
We need to update to reduce possible issues with downgrading until we
fix the behavior of setuptools search for setup requirements from
pygradle.


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
```diff
@@ -65,7 +65,7 @@ class PythonExtension {
 
     /** A way to define forced versions of libraries */
     public Map<String, Map<String, String>> forcedVersions = [
-        'appdirs'       : ['group': 'pypi', 'name': 'appdirs',          'version': '1.4.0'],
```
ethankhall: You also need to update the importer to include this version.

zvezdan: Done!

PR #133, opened 2017-05-15
Change installation order to post-order graph traversal.

zvezdan:
Add the new utility that resolves the dependencies in the post-order.
Use it to get a file collection and apply that collection to the
PipInstallTask and BuildWheelsTask for all configurations.

Now we can drop pbr from setup requirements.

Added comprehensive unit tests for the utility class.

Added additional assertions in an existing integration test.
Checked that it passed with expectation for sorted build dependencies on
master and that it failed with the same expectation with this change.
Once the condition was negated, it passed with this change.

Bumped the version to 0.5.0.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
@@ -67,7 +67,6 @@ class PythonExtension {
     public Map<String, Map<String, String>> forcedVersions = [
         'argparse'      : ['group': 'pypi', 'name': 'argparse',         'version': '1.4.0'],
         'flake8'        : ['group': 'pypi', 'name': 'flake8',           'version': '2.5.4'],
-        'pbr'           : ['group': 'pypi', 'name': 'pbr',              'version': '1.8.0'],
         'pex'           : ['group': 'pypi', 'name': 'pex',              'version': '1.1.4'],
         'pip'           : ['group': 'pypi', 'name': 'pip',              'version': '7.1.2'],
         'pytest'        : ['group': 'pypi', 'name': 'pytest',           'version': '2.9.1'],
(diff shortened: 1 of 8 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
```diff
@@ -67,15 +67,13 @@ class PythonExtension {
     public Map<String, Map<String, String>> forcedVersions = [
         'argparse'      : ['group': 'pypi', 'name': 'argparse',         'version': '1.4.0'],
         'flake8'        : ['group': 'pypi', 'name': 'flake8',           'version': '2.5.4'],
-        'pbr'           : ['group': 'pypi', 'name': 'pbr',              'version': '1.8.0'],
```
ethankhall: What magic lets us remove pbr?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/DependencyOrder.java
```diff
@@ -0,0 +1,127 @@
...
+     * @param root the root of the dependency (sub)tree
+     * @param seen the input set of already seen dependencies
+     * @param dependencies the output set of dependencies in post-order
+     * @return the root itself
+     */
+    public static ResolvedComponentResult postOrderDependencies(
+        ResolvedComponentResult root,
+        Set<ComponentIdentifier> seen,
+        Set<ResolvedComponentResult> dependencies
+    ) {
```
ethankhall: Move this onto the same line.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/PythonExtension.groovy
zvezdan: The fact that we install in graph post-order.

Now when a package that really has pbr in `setup_requires` gets installed, pbr will sniff the virtualenv, but all the required dependencies will already be there and pbr will be happy.

With previous alphabetical order, most of the dependencies would be installed late (e.g., psutils) although they are needed early in the dependency graph. In such cases, pbr would immediately bail out complaining about missing dependencies. The workaround was to install pbr itself early when virtualenv has only base setup requirements and no missing dependencies. The downside was that we could not upgrade setuptools beyond 19.1.1 and pbr was artificially made a dependency of **every** package.

Now, we can remove it, and it will be brought in on-demand only for packages that really need it. Also, we can finally upgrade our setup requirements (setuptools, virtualenv, ...) to support Python 3.6 properly.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/DependencyOrder.java
zvezdan: They don't fit in 120 characters margin.
That's why I put them on separate lines.

ethankhall: My comment was just focused on the `) {`

zvezdan: Done!

PR #145, opened 2017-06-16
Upgrade setup and build requirements for Python 3.6.

zvezdan:
With previously committed changes we now have everything in place for
upgrade of setuptools, virtualenv, pip, and pex. We also upgrade pytest
and flake8 to match versions used internally at LinkedIn.

Also, bumped the version to 0.6.0 and removed the issue with pip-7
because we now use pip-9 with our patch applied
(https://github.com/pypa/pip/pull/3701).

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,13 +93,3 @@ PyGradle works on the following platforms.
 Additional compatibility notes are availabe [here](docs/compatibility.md).
 We're happy to review and merge pull requests that add support for additional
 platforms.
-
-# Known Potential Issues
-
-- Due to a bug in pip, when trying to install scipy may fail. A potential work
-around is to use a newer version of pip. A PR was merged into pip master that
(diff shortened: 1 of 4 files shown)


zvezdan:
@sixninetynine Two issues with the columns.

1. They depend on the setup in IDE -- this setup is off in Java and on in Groovy by default -- I use the same setting for both.
2. Whenever you add a new entry that is longer than the currently longest key, the whole mapping gets re-aligned. So, instead of a one line diff it looks like all the lines changed -- as it does here.

I'd rather have this change once and one-line changes in the future, then every time trying to read into diff and find out what actually changed and what was just re-aligned.

That's why Python style avoids alignments as do many other styles including Java -- both externally and internally at LinkedIn.

I think you should change your Groovy setting in the IDE instead. :-)

PR #149, opened 2017-06-27
Always install snapshot versions. Fixes #135

### LINKED ISSUE #135: PipInstallTask should always install SNAPSHOT versions
Since we use gradle's inputs/outputs to determine if some package must be pip installed(or upgraded), we should use --upgrade flag to turn off pip's check which is only based on the version

This is helpful on incremental builds where we are updating snapshots/same-versions of dependencies. In such case, gradle runs the pip install task as expected, but pip thinks the dependency is already satisfied:

`Requirement already satisfied (use --upgrade to upgrade):`

schakrava:
SNAPSHOT versions of dependencies(created by gradle) are possibly different every time. So when pip installing packages ending with -SNAPSHOT, use
--ignore-installed.


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PipInstallTask.groovy
```diff
@@ -155,18 +155,23 @@ class PipInstallTask extends DefaultTask implements FailureReasonProvider {
...
                                    '--disable-pip-version-check',
                                    '--no-deps']
                 commandLine.addAll(args)
-                commandLine.add(installable.getAbsolutePath())
 
-                if (project.file(egg).exists() || project.file(dist).exists()) {
-                    logger.lifecycle(PythonHelpers.createPrettyLine("Install ${shortHand}", "[SKIPPING]"))
-                    continue
+                if (shortHand.endsWith('-SNAPSHOT')) {
+                    /* snapshot packages may have changed, so reinstall them every time */
```
zvezdan: Use `// ...` for single line comments, please.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PipInstallTask.groovy
```diff
@@ -155,18 +155,23 @@ class PipInstallTask extends DefaultTask implements FailureReasonProvider {
...
                                    'install',
                                    '--disable-pip-version-check',
                                    '--no-deps']
                 commandLine.addAll(args)
-                commandLine.add(installable.getAbsolutePath())
 
-                if (project.file(egg).exists() || project.file(dist).exists()) {
-                    logger.lifecycle(PythonHelpers.createPrettyLine("Install ${shortHand}", "[SKIPPING]"))
-                    continue
+                if (shortHand.endsWith('-SNAPSHOT')) {
```
zvezdan: Should this go before line 168 `addAll(args)`?

schakrava: no, as long as it's added before the installable it's good.

PR #166, opened 2017-09-13
log warning message with missing resolved artifact information.

### LINKED ISSUE #165: warn when a resolved artifact is missing
When collecting configuration-files/dependencies in post-order, print an explicit warning with the missing resolved artifact information. This helps the user pinpoint the reason why we fallback to sorted order.

schakrava:
It's useful to the user to know which artifact is unresolved that causes
fallback from post-order to sorted order

Fixes #165

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/DependencyOrder.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/DependencyOrder.java
@@ -32,12 +32,17 @@ import org.gradle.api.artifacts.component.ComponentIdentifier;
 import org.gradle.api.artifacts.result.DependencyResult;
 import org.gradle.api.artifacts.result.ResolvedComponentResult;
 import org.gradle.api.artifacts.result.ResolvedDependencyResult;
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
 
 
 /**
  * A container of static utilities for dependency order.
  */
 public class DependencyOrder {
+
+    private static final Logger logger = Logging.getLogger(DependencyOrder.class);
+
     private DependencyOrder() {
         // prevent instantiation of the utility class
     }
@@ -114,7 +119,11 @@ public class DependencyOrder {
 
         // Prepare an ordered set of files in post-order.
         for (ResolvedComponentResult d : configurationPostOrderDependencies(configuration)) {
-            files.add(idToFileMap.get(d.getModuleVersion()));
+            if (!idToFileMap.containsKey(d.getModuleVersion())) {
+                logger.warn("***** WARNING Missing resolved artifact for " + d);
+            } else {
+                files.add(idToFileMap.get(d.getModuleVersion()));
+            }
         }
 
         // Make sure the files set is a subset of configuration files.


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/DependencyOrder.java
```diff
@@ -114,7 +119,11 @@ public static ResolvedComponentResult postOrderDependencies(
 
         // Prepare an ordered set of files in post-order.
         for (ResolvedComponentResult d : configurationPostOrderDependencies(configuration)) {
-            files.add(idToFileMap.get(d.getModuleVersion()));
+            if (!idToFileMap.containsKey(d.getModuleVersion())) {
+                logger.lifecycle("***** WARNING Missing resolved artifact for " + d);
```
ethankhall: Move this up to warning, if it's actually a warning.

schakrava: Thanks @ethankhall . pr updated

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/DependencyOrder.java
```diff
@@ -114,7 +119,11 @@ public static ResolvedComponentResult postOrderDependencies(
 
         // Prepare an ordered set of files in post-order.
         for (ResolvedComponentResult d : configurationPostOrderDependencies(configuration)) {
-            files.add(idToFileMap.get(d.getModuleVersion()));
```
zvezdan: Why are we doing this conditionally now?
This is not the same behavior.

zvezdan: To clarify ...
The `files.add` should be unconditional just as it was before.
There's no need for `else`.

schakrava: without the else, it would just add a `null` to files for missing artifacts. Is that helpful? I found it to be confusing while debugging an issue.

zvezdan: I believe that we need to investigate deeper the reasons for this behavior then.
The files in resolved configuration whould be in the map.
If they are not, we have a different problem than just reporting the missing file.
I'd like to see us get to the root of this instead of just scratching the surface.

zvezdan: If this is urgent, feel free to proceed because I won't be able to respond until tomorrow.
But please put a TODO to investigate this further if you do that.

schakrava: This statement "The files in resolved configuration should be in the map." is not true if an artifact cannot be resolved. This case is reproducible when dependencies are improperly defined in case of multi-variant builds, for example. There's an internal ticket on which you are added as a watcher. Please see that for more context.

I don't think that the current behavior is buggy, but 1) there's no use in adding a null to `files` which the if statement prevents and 2) warning the user about the artifact that was not resolved is actionable compared to just throwing a non-specific exception message.

This is not blocking anything, so it's not urgent IMO. Feel free to investigate more and approve when you are satisfied.

zvezdan: Please commit the change.
That warning is useful for diagnostics.

The key here is "... when dependencies are improperly defined in case of multi-variant builds ..." which confirmed my suspicion that something deeper unexpected and **unrelated to PyGradle** is going on. I agree that this is very specific to an internal use-case and will have to be solved internally.

schakrava: Thanks @zvezdan

PR #176, opened 2017-12-18
Pygradle will build wheels first

ethankhall:
This change has a few components:
 - Upgrade to Java 8
 - Upgrade to Gradle 4.4
 - Make Pygradle build wheels, if possible, before installs.

## Upgrade to Java 8
Java 7 is EOL and no longer receiving updates. https://www.java.com/en/download/faq/java_7.xml

This upgrade will allow us to replace Groovy code more effectively with Java code. This should make sure we don't introduce incompatibilities.

Several classes were migrated from Groovy to Java.

## Upgrade to Gradle 4.4
Why not? We are now using an internal API, but it works on the 4.x series (tested). We should update the recommended minimum to 4.x.

## Build Wheels First
In an effort to make Python builds faster, we should build wheels and shove them into a global cache. This global cache will make builds faster after the first build. It also allows builds across multiple projects to reuse work that has already been done.

From some internal tests, I've seen as much as a 20min decrease in build time (when the wheels were already available).

This change also includes a bunch of output changes. Mainly, we will stop spamming the console with output and use Gradle's internal API (and move to a public one when available) to update the status line, so we don't need to show new lines on the console so the user know's that the process is still running.


pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PythonPluginIntegrationTest.groovy
```diff
@@ -122,30 +125,28 @@ class PythonPluginIntegrationTest extends Specification {
...
+        !installOrderSorted(result.output, ':foo:installBuildRequirements', ':foo:installPythonRequirements')
     }
 
     def installOrderSorted(String output, String start, String end) {
-        def lines = output.split(/\n/)
+        println output
+        def begining = output.indexOf(start)
+        def ending = output.lastIndexOf(end)
+        def lines = output.substring(begining, ending).readLines()
+        println lines
```
mockitoguy: Please remove the debug statements. Thank you!

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/wheel/WheelCacheTest.groovy
```diff
@@ -0,0 +1,93 @@
...
+        temporaryFolder.newFile('lib-1.2-' + wheelName + ".whl")
+
+        def cache = new WheelCache(temporaryFolder.root, pythonTag, platformTag)
+
+        expect:
+        cache.findWheel('lib', '1.2', pythonVersion).isPresent() == matches
+
+        where:
+        wheelName                        | pythonVersion | pythonTag | platformTag | matches
+        'cp33-cp33m-linux_x86_64'        | PYTHON_33     | C_PYTHON  | LINUX_TAG   | true
```
zvezdan: We don't use Python 3.3 and it's not supported any more.
3.4 is probably have the same destiny too.
We use 3.5 and 3.6 internally.
Could we make test examples with Python 3.5 and 3.6 only?

zvezdan: We don't use Python 3.3 and it's not supported any more.
3.4 is probably have the same destiny too.
We use 3.5 and 3.6 internally.
Could we make test examples with Python 3.5 and 3.6 only?

ethankhall: Those values came from https://www.python.org/dev/peps/pep-0425/#id1. I would prefer to use the values from the PEP. We also need to make sure we have different major versions.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildPexTask.java
```diff
@@ -121,6 +125,9 @@ private void configureExecution(PythonExtension pythonExtension, ExecSpec spec)
             wheelExtension.getWheelCache().getAbsolutePath(),
             "--no-deps",
             ".");
+        spec.setErrorOutput(outputStream);
+        spec.setErrorOutput(outputStream);
```
zvezdan: Should one of these two lines be `setStandardOutput` instead of both setting error output?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
```diff
@@ -114,17 +124,37 @@ class BuildWheelsTask extends DefaultTask {
...
+
+        int counter = 0
         installables.each { File installable ->
 
             def packageInfo = PackageInfo.fromPath(installable.path)
             def shortHand = packageInfo.version ? "${packageInfo.name}-${packageInfo.version}" : packageInfo.name
-            def messageHead = 'Preparing wheel ' + shortHand
+
+            def clock = taskTimer.start(shortHand)
+            progressLogger.progress("Preparing wheel $shortHand (${++counter} of ${installables.size()})")
```
zvezdan: The `installables.size()` should be stored into a variable before this loop instead of calling it repeatedly here.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
```diff
@@ -114,17 +124,37 @@ class BuildWheelsTask extends DefaultTask {
...
+
+            def wheel = wheelCache.findWheel(packageInfo.name, packageInfo.version, pythonExtension.details.getPythonVersion())
```
zvezdan: Let's put this below the `if packageExcludeFilter ...` below to avoid searching for the wheel if the package is going to be excluded anyway.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
```diff
@@ -114,17 +124,37 @@ class BuildWheelsTask extends DefaultTask {
...
+            if (wheel.isPresent()) {
+                FileUtils.copyFile(wheel.get(), new File(wheelExtension.wheelCache, wheel.get().name))
+                return
+            }
 
             if (packageExcludeFilter.isSatisfiedBy(packageInfo)) {
-                logger.lifecycle(PythonHelpers.createPrettyLine(messageHead, "[EXCLUDED]"))
```
zvezdan: Do we still need logging on some level of these packages that have been excluded so that we can debug/reason about them and whether the exclusion was correct?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/ParallelWheelGenerationTask.java
```diff
@@ -0,0 +1,166 @@
...
+            PackageInfo packageInfo = PackageInfo.fromPath(file.getPath());
+            TaskTimer.TickingClock clock = taskTimer.start(packageInfo.getName() + "-" + packageInfo.getVersion());
+            makeWheelFromSdist(progressLogger, totalSize, wheelCache, file);
+            clock.stop();
+        });
+
+        try {
+            FileUtils.write(getBuildReport(), taskTimer.buildReport());
+        } catch (IOException ignore) {
+        }
```
zvezdan: Can you add a comment inside this block, such as `// PASS`
Similar for other instances of this pattern.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/ParallelWheelGenerationTask.java
```diff
@@ -0,0 +1,166 @@
...
+        progressLogger.progress(String.format("Building wheel %s %d of %d", packageInfo.getName(), counter.incrementAndGet(), totalSize));
+        Optional<File> cachedWheel = wheelCache.findWheel(packageInfo.getName(), packageInfo.getVersion(), pythonDetails.getPythonVersion());
+
+        if (cachedWheel.isPresent()) {
+            return;
+        }
+
+        ByteArrayOutputStream stream = new ByteArrayOutputStream();
+        ExecResult results = getProject().exec(exec -> {
+            exec.commandLine(pythonDetails.getVirtualEnvInterpreter(),
```
zvezdan: Where do we set environment for this command execution?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/ParallelWheelGenerationTask.java
```diff
@@ -0,0 +1,166 @@
...
+        WheelCache wheelCache = new WheelCache(cacheDir, pythonTag, platformTag);
+
+        progressLogger.started();
+
+        TaskTimer taskTimer = new TaskTimer();
+
+        Set<File> files = getFilesToConvert().getFiles();
+        int totalSize = files.size();
+        files.parallelStream().forEach(file -> {
+            PackageInfo packageInfo = PackageInfo.fromPath(file.getPath());
```
zvezdan: How do I exclude certain dependency from this generic wheel building.
Let's assume I have several projects checked out.
For all but one of them I need a wheel for package xxx. However, for one of them I need a custom built xxx with different options or environment.
The existing wheel in the cache would not be satisfactory.
I need a way to skip using a generic wheel here and rebuild from sdist for that project
I don't currently see a way to do that with this approach.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PipInstallTask.groovy
```diff
@@ -134,81 +142,106 @@ class PipInstallTask extends DefaultTask implements FailureReasonProvider {
...
-                File egg = sitePackages.resolve("${sanitizedName}-${packageInfo.version}-py${pyVersion}.egg-info").toFile()
-                File dist = sitePackages.resolve("${sanitizedName}-${packageInfo.version}.dist-info").toFile()
-
-                def mergedEnv = new HashMap(extension.pythonEnvironment)
-                if (environment != null) {
-                    mergedEnv.putAll(environment)
-                }
-
-                if (project.file(egg).exists() || project.file(dist).exists()) {
-                    logger.lifecycle(PythonHelpers.createPrettyLine("Install ${shortHand}", "[SKIPPING]"))
```
zvezdan: Do we still need this logging on some level to be able to debug/reason about it?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/DependencyOrder.java
```diff
@@ -35,6 +27,14 @@
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
 
+import java.io.File;
```
zvezdan: This file has no other change except this. Did we even need to change it then?
What if org.gradle.api.... has a Collection class? It should be able to override the java.util.Collection. But it cannot because of this import order.

I'm not saying it's a good idea. I'm just saying the original order seems correct to me.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/NoopOutputStream.java
```diff
@@ -13,24 +13,13 @@
...
-    target.plugins.apply(PythonPlugin)
-    settings = project.getExtensions().getByType(PythonExtension)
-    applyTo(target)
-  }
-
-  abstract void applyTo(Project project)
+public class NoopOutputStream extends OutputStream {
+    @Override
+    public void write(int b) {
+        //noop
```
zvezdan: Can you please add one space between the comment signs and the text: `// noop`.
Similarly for other instances of this pattern in this patch.
You can find them by searching for `//n` on the review page in the browser.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/internal/pex/PexExecOutputParser.java
```diff
@@ -15,14 +15,14 @@
...
 
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
 import org.gradle.api.GradleException;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
 import org.gradle.process.ExecResult;
 
+import java.util.regex.Matcher;
```
zvezdan: Again, I don't see the need for this change.
We can talk again about import order in case I'm misunderstanding something.

Ditto for the next file, except, in it, an empty line should be moved below java.util.Map import. So, perhaps it's OK to change that one.

There are other places like this. That's why I'm not commenting on them, until we figure out the intent.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/internal/pex/ThinPexGenerator.java
```diff
@@ -45,26 +44,26 @@
...
 
     public ThinPexGenerator(
-            Project project,
-            List<String> pexOptions,
-            EntryPointTemplateProvider templateProvider,
-            Map<String, String> extraProperties) {
+        Project project,
+        List<String> pexOptions,
+        EntryPointTemplateProvider templateProvider,
+        Map<String, String> extraProperties) {
```
zvezdan: This line (and other arguments above it) was indented more in the original (IDE default I believe) to be distinguishable from the first line of the method code that follows next. So, I'm not sure about the usefulness of this dedent.

Again, something where we need to figure out what's the IDE default.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/pip/PipFreezeOutputParser.java
```diff
@@ -15,11 +15,12 @@
  */
 package com.linkedin.gradle.python.util.pip;
 
+import org.gradle.api.GradleException;
```
zvezdan: Another unnecessary change.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/wheel/WheelCache.java
```diff
@@ -0,0 +1,90 @@
...
+            .filter(file -> pythonAbi.stream().anyMatch(it -> file.getName().contains(it)))
+            .filter(file -> platformTagList.stream().anyMatch(it -> file.getName().endsWith(it)))
+            .collect(Collectors.toList());
+
+        logger.debug("Found artifacts: {}", collect);
+
+        if (collect.isEmpty()) {
+            return Optional.empty();
+        } else {
+            return Optional.of(collect.get(0));
```
zvezdan: What does this return?
Some comments on this class would be nice (and other new classes too).

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/wheel/WheelCacheTest.groovy
zvezdan: OK. Can we add some Python 2.7 tests here then. Notice that the format will be `macosx-10.12-intel` instead of `macosx-10.10-x86_64` for the command used to produce that string. We need to check if there's the similar difference on Linux.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PlatformTag.java
```diff
@@ -0,0 +1,66 @@
...
+    public String getPlatform() {
+        return platform;
+    }
+
+    public static PlatformTag makePlatformTag(Project project, PythonDetails pythonDetails) {
+        try (ByteArrayOutputStream stream = new ByteArrayOutputStream()) {
+
+            List<String> args = Arrays.asList(pythonDetails.getVirtualEnvInterpreter().getAbsolutePath(),
+                "-c",
+                "import distutils; print(distutils.util.get_platform())");
```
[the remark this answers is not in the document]
ethankhall: The value needs to be consistent with what the wheel has in it's filename. Fixed the import.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PlatformTag.java
```diff
@@ -0,0 +1,66 @@
...
+            return new PlatformTag(platform);
+
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    @Override
+    public String toString() {
+        return "PlatformTag{"
```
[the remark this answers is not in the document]
ethankhall: This is auto-generated code by IntelliJ.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PlatformTag.java
```diff
@@ -0,0 +1,66 @@
...
+            + "platform='" + platform + '\''
```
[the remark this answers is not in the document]
ethankhall: This is auto-generated code by IntelliJ. I would prefer to leave it like it is.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonTag.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class PythonTag implements Serializable {
+    public static final PythonTag C_PYTHON = new PythonTag("cp", "cpython");
+    public static final PythonTag IRON_PYTHON = new PythonTag("ip", "ironpython");
+    public static final PythonTag PY_PY = new PythonTag("pp", "pypy");
+    public static final PythonTag JYTHON = new PythonTag("jy", "python");
```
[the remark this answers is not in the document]
ethankhall: Yep good catch!

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonVersionParser.java
```diff
@@ -0,0 +1,51 @@
...
+            execSpec.setErrorOutput(outputStream);
+            execSpec.setIgnoreExitValue(true);
+        });
+
+        String output = outputStream.toString();
+        if (execResult.getExitValue() != 0) {
+            throw new GradleException(output);
+        }
+
+        String versionString = output.split(" ")[1];
```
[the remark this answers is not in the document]
ethankhall: I don't know, I'll add it back.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/PythonHelpers.java
```diff
@@ -13,26 +13,31 @@
...
+    public static final int LINE_WIDTH = 80;
 
-    public static final int LINE_WIDTH = 80
+    private PythonHelpers() {
+        //NOOP
+    }
 
     /**
      * Detect if we're connected to a TTY and support color.
      * <p>
```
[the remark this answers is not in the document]
ethankhall: Yeah. Auto-generated in the conversion.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/PythonHelpers.java
```diff
@@ -41,43 +46,47 @@ protected static boolean isTty(Project project) {
...
      */
     public static String successFlair(Project project, PythonExtension settings) {
 
-        StringBuilder successFlair = new StringBuilder()
+        StringBuilder successFlair = new StringBuilder();
 
-        if (settings.consoleOutput != ConsoleOutput.RAW) {
-            successFlair << ColorHelper.ANSI_GREEN
+        if (!settings.consoleOutput.equals(ConsoleOutput.RAW)) {
+            StringGroovyMethods.leftShift(successFlair, ColorHelper.ANSI_GREEN);
```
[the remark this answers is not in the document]
ethankhall: Groovy to Java conversion. I missed it.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/PythonPlugin.groovy
```diff
@@ -1,340 +0,0 @@
...
-        def packageJsonDocsTask = project.tasks.create(TASK_PACKAGE_JSON_DOCS.value, Tar, new PackageDocumentationAction(sphinxJson))
-        packageJsonDocsTask.dependsOn(project.tasks.getByName(TASK_BUILD_DOCS.value))
-
-        // Only build the artifact if the docs source directory actually exists
-        if (project.file(settings.docsDir).exists()) {
-            project.artifacts.add(CONFIGURATION_PYDOCS.value, packageDocsTask)
-            project.artifacts.add(CONFIGURATION_PYDOCS.value, packageJsonDocsTask)
-        }
-
-        project.tasks.create(TASK_SETUP_PY_WRITER.value, GenerateSetupPyTask)
```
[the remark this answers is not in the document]
ethankhall: Omission. Fixed.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/PythonPlugin.java
```diff
@@ -0,0 +1,152 @@
...
+import static com.linkedin.gradle.python.util.StandardTextValues.CONFIGURATION_TEST;
+import static com.linkedin.gradle.python.util.StandardTextValues.CONFIGURATION_VENV;
+import static com.linkedin.gradle.python.util.StandardTextValues.CONFIGURATION_WHEEL;
+import static com.linkedin.gradle.python.util.StandardTextValues.TASK_CLEAN_SAVE_VENV;
+import static com.linkedin.gradle.python.util.StandardTextValues.TASK_PIN_REQUIREMENTS;
+import static com.linkedin.gradle.python.util.StandardTextValues.TASK_SETUP_LINKS;
+import static com.linkedin.gradle.python.util.StandardTextValues.TASK_VENV_CREATE;
+
+public class PythonPlugin implements Plugin<Project> {
+    public void apply(final Project project) {
```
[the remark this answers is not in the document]
ethankhall: After Java 6 you don't need `@Override` anymore.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/PythonPlugin.java
```diff
@@ -0,0 +1,152 @@
...
+        createConfigurations(project);
+        configureVendedDependencies(project, settings);
+
+        /*
+         * To prevent base dependencies, such as setuptools, from installing/reinstalling, we will
+         * pin their versions to values in the extension.forcedVersions map, which will contain known
+         * good versions that satisfy all the requirements.
+         */
+        final PyGradleDependencyResolveDetails dependencyResolveDetails = new PyGradleDependencyResolveDetails(settings.forcedVersions);
+        //noinspection GroovyAssignabilityCheck
```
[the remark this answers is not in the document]
ethankhall: Nope. Missed it.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/PythonWheelDistributionPlugin.java
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+package com.linkedin.gradle.python.plugin;
+
+import com.linkedin.gradle.python.util.StandardTextValues;
+import org.gradle.api.Project;
+
+import java.io.File;
+import java.util.LinkedHashMap;
+
```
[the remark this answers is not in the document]
ethankhall: I'm just going to delete it.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/WheelFirstPlugin.java
```diff
@@ -0,0 +1,77 @@
...
+                    .plus(configurations.getByName(CONFIGURATION_VENV.getValue()))
+                    .plus(configurations.getByName(CONFIGURATION_TEST.getValue()));
+
+                it.setFilesToConvert(dependencies);
+                it.setCacheDir(cacheDir);
+                it.dependsOn(tasks.getByName(TASK_INSTALL_SETUP_REQS.getValue()));
+            });
+
+            tasks.withType(PipInstallTask.class, it -> {
+                if (!Objects.equals(it.getName(), TASK_INSTALL_SETUP_REQS.getValue())) {
```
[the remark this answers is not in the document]
ethankhall: It is a `PipInstallTask`.. What string comparison are you thinking?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/internal/InstallDependenciesPlugin.java
```diff
@@ -0,0 +1,112 @@
...
+         */
+        project.getTasks().create(StandardTextValues.TASK_INSTALL_PROJECT.getValue(), PipInstallTask.class, it -> {
+            it.setPythonDetails(settings.getDetails());
+            it.dependsOn(project.getTasks().getByName(StandardTextValues.TASK_INSTALL_PYTHON_REQS.getValue()));
+            it.setInstallFileCollection(project.files(project.file(project.getProjectDir())));
+            it.setArgs(Collections.singletonList("--editable"));
+            it.setEnvironment(settings.pythonEnvironmentDistgradle);
+        });
+
+        /*
```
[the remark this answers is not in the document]
ethankhall: I kept TASK_INSTALL_PROJECT, and moved the rest to the other plugin.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildPexTask.java
ethankhall: Yep. Fixed.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
ethankhall: Added ```if (pythonExtension.consoleOutput == ConsoleOutput.RAW) {
                LOGGER.lifecycle("Installing {}", shortHand)
            }
``` that should happen on CI machines.

File: pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/ParallelWheelGenerationTask.java (hunk 9 above)
Comment (ethankhall) [reply pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/ParallelWheelGenerationTask.java]: You don't. We would be a generic on first, then build a different once second. If the wheel can't be built, it will fail, and ignore silently. The pip-install will then fallback to doing it the old way.

File: pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/ParallelWheelGenerationTask.java (hunk 8 above)
Comment (ethankhall) [reply pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/ParallelWheelGenerationTask.java]: We don't.

File: pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/DependencyOrder.java (hunk 11 above)
Comment (ethankhall) [reply pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/DependencyOrder.java]: Default IntelliJ does this formatting. When I was fixing imports it happened.

File: pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/internal/pex/ThinPexGenerator.java (hunk 14 above)
Comment (ethankhall) [reply pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/internal/pex/ThinPexGenerator.java]: I have a clean install of IntelliJ. We can apply a strict style guide after this PR and make it automated.

**Comment by dsully** [comment]
Would love to see some traction on this. Ethan, is it possible to break your change up?

**Comment by ethankhall** [comment]
@dsully I asked @zvezdan if he wanted it split, and he said it was ok as is. I have a branch that has the Java changes separate if we need to split it out.

I recently was found and fixed a bug that was causing problems with the testing. We have one open question that @zvezdan and I need to worth through before pushing. Once that's done it should be good to commit it.

File: pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/WheelFirstPlugin.java [hunk 30]
```diff
@@ -0,0 +1,86 @@
...
+
+        File cacheDir = new File(project.getGradle().getGradleUserHomeDir(), "pygradle-wheel-cache");
+
+        project.getPlugins().withType(PythonPlugin.class, plugin -> {
+
+            SupportedWheelFormats supportedWheelFormats = new SupportedWheelFormats();
+            CachedBackedWheelCache wheelCache = new CachedBackedWheelCache(cacheDir, supportedWheelFormats);
+
+            FindAbiForCurrentPythonTask abiScript = tasks.create("findPythonAbi", FindAbiForCurrentPythonTask.class, it -> {
+                it.dependsOn(tasks.getByName(TASK_INSTALL_SETUP_REQS.getValue()));
```
zvezdan: This should be able to execute as soon as `TASK_VENV_CREATE` finishes. No need to wait this long for setup requires task to finish that has much more dependencies. The `wheel` package is vended inside virtualenv.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/WheelFirstPlugin.java
```diff
@@ -0,0 +1,86 @@
...
+                it.setSupportedWheelFormat(supportedWheelFormats);
+                it.addPythonDetails(() -> ExtensionUtils.getPythonExtension(project).getDetails());
```
zvezdan: How does this work for builds with multiple-versions of python?
Is this currently used for sdist builds? If yes, then it needs to pass details into a task -- similar to `PipInstallTask`.
If this is only used for deployables, then fetching the main extension details as done here is fine (one Python version build).

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/WheelFirstPlugin.java
```diff
@@ -0,0 +1,86 @@
...
+                    .plus(configurations.getByName(CONFIGURATION_PYTHON.getValue()))
+                    .plus(configurations.getByName(CONFIGURATION_WHEEL.getValue()))
+                    .plus(configurations.getByName(CONFIGURATION_PYDOCS.getValue()))
+                    .plus(configurations.getByName(CONFIGURATION_VENV.getValue()))
+                    .plus(configurations.getByName(CONFIGURATION_TEST.getValue()));
+
+                it.setFilesToConvert(dependencies);
+                it.setWheelCache(wheelCache);
+                it.setCacheDir(cacheDir);
+                it.dependsOn(tasks.getByName(TASK_INSTALL_SETUP_REQS.getValue()));
```
zvezdan: Could these two also be `TASK_VENV_CREATE` instead of waiting for setup requirements?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
```diff
@@ -114,17 +127,37 @@ class BuildWheelsTask extends DefaultTask {
...
+            def clock = taskTimer.start(shortHand)
+            progressLogger.progress("Preparing wheel $shortHand (${ ++counter } of $numberOfInstallables)")
+
+            if (pythonExtension.consoleOutput == ConsoleOutput.RAW) {
+                LOGGER.lifecycle("Installing {}", shortHand)
+            }
 
             if (packageExcludeFilter.isSatisfiedBy(packageInfo)) {
-                logger.lifecycle(PythonHelpers.createPrettyLine(messageHead, "[EXCLUDED]"))
+                return
```
zvezdan: Having excluded packages logged is really helpful for reasoning about conditional dependencies between Python 2 and 3 if anything goes wrong with the build. Can we get the logging back in?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/ParallelWheelGenerationTask.java
zvezdan: Can you be a little more specific?
Here's an example. I want to install the package `Pillow` but pass to it option `--disable-tiff`. The generic will with tiff enabled will not cut it for me.

Can I prevent the build of generic Pilllow wheel at all?

If not, can I at least get an override build with my option, assuming this generic build happens before the regular dependency install and my regular dependency install takes care to ignore the generic wheel?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PipInstallTask.groovy
zvezdan: Can you comment on this old question above?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PipInstallTask.groovy
```diff
@@ -134,81 +141,105 @@ class PipInstallTask extends DefaultTask implements FailureReasonProvider {
...
+        progressLogger.completed()
+
+        new File(project.buildDir, getName() + "-task-runtime-report.txt").text = taskTimer.buildReport()
+    }
+
+    @SuppressWarnings("ParameterCount")
+    private void doInstall(String shortHand, PackageInfo packageInfo, Path sitePackages,
+                           String pyVersion, PythonExtension extension, File installable) {
+        if (packageExcludeFilter.isSatisfiedBy(packageInfo)) {
+            return
```
zvezdan: Loggin of excluded packages really helps with reasoning about the build failure for conditional dependencies between Python 2 and 3. Can we put the logging back please?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/DependencyOrder.java
```diff
@@ -49,7 +49,7 @@ private DependencyOrder() {
 
     /**
      * Traverses the dependency tree post-order and collects dependencies.
-     * <p>
+     *
```
zvezdan: See the comment above about `<p>` in Javadoc.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/StandardTextValues.groovy
```diff
@@ -20,7 +20,7 @@ package com.linkedin.gradle.python.util
  * are being moved to this enum.  Its an Enum rather than an interface because according to CodeNarc, interfaces of nothing
  * but constants is now taboo.
  */
-enum StandardTextValues {
+public enum StandardTextValues {
```
zvezdan: This is a Groovy file. It's public by default and IDE will warn about unnecessary qualifier. If you want to make it Java, though, this would be needed.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/wheel/SupportedWheelFormats.java
```diff
@@ -0,0 +1,63 @@
...
+            .anyMatch(it -> contains(it, pythonTags, abiTags, platformTags));
+    }
+
+    private static boolean contains(AbiTriple triple, String[] pythonTags, String[] abiTags, String[] platformTags) {
+        return contains(triple.getPythonTag(), pythonTags)
+          && contains(triple.getAbiTag(), abiTags)
+          && contains(triple.getPlatformTag(), platformTags);
+    }
+
+    private static boolean contains(String neadle, String[] haystack) {
```
zvezdan: Spelling typo. Please change `neadle` to `needle`. You're looking for a "needle" in a "haystack".

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonTag.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public String getPrefix() {
+        return prefix;
+    }
+
+    static public PythonTag findTag(Project project, PythonDetails pythonDetails) {
+        List<String> args = new ArrayList<>(Arrays.asList(pythonDetails.getVirtualEnvInterpreter().getAbsolutePath(), "-c"));
+
+        if ("2".equals(pythonDetails.getPythonVersion().getPythonMajor())) {
```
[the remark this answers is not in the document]
kopy07: Or I could delete it :-)

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PlatformTag.java
```diff
@@ -0,0 +1,66 @@
...
+    public String getPlatform() {
+        return platform;
+    }
+
+    public static PlatformTag makePlatformTag(Project project, PythonDetails pythonDetails) {
+        try (ByteArrayOutputStream stream = new ByteArrayOutputStream()) {
+
+            List<String> args = Arrays.asList(pythonDetails.getVirtualEnvInterpreter().getAbsolutePath(),
+                "-c",
+                "import distutils.util; print(distutils.util.get_platform())");
```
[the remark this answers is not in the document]
kopy07: I forgot to clean up the code. This class should be deleted.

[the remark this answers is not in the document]
ethankhall: I can actually delete this and the PythonTag class as they aren't needed anymore.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/PythonFlyerPlugin.java
```diff
@@ -13,36 +13,37 @@
...
+import org.gradle.api.tasks.Copy;
 
-import java.nio.file.Paths
+import java.io.File;
+import java.nio.file.Paths;
 
 /**
  * A Flyer plugin.
- * <p>
+ *
```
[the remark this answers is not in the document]
ethankhall: Fixed.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PlatformTag.java
```diff
@@ -0,0 +1,66 @@
...
+import org.gradle.api.Project;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.Serializable;
+import java.io.UncheckedIOException;
+import java.util.Arrays;
+import java.util.List;
+
+public class PlatformTag implements Serializable {
```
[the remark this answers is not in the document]
ethankhall: Class was removed.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/PythonTag.java
```diff
@@ -0,0 +1,88 @@
...
+import org.gradle.api.Project;
+import org.gradle.process.ExecResult;
+
+import java.io.ByteArrayOutputStream;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+// Implementation of https://www.python.org/dev/peps/pep-0425
```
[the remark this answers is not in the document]
ethankhall: Class was removed.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PipInstallTask.groovy
ethankhall: Done.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PipInstallTask.groovy
ethankhall: Yes. Otherwise in CI it' wont' show up, so we won't be able to debug what happened.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/pip/PipFreezeOutputParser.java
ethankhall: This has been fixed I believe. (Only one import exists now).

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/plugin/WheelFirstPlugin.java
kopy07: For each version of python, you would call `it.addPythonDetails(() -> ...);` Each version of python will be checked for it's ABI options and those are stored against the path to the venv interrupter.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/ParallelWheelGenerationTask.java
kopy07: At the moment you can't. When your in-progress change lands (package specific options) they would be excluded from the wheel generation and from being picked up in the wheel options.

PR #185, opened 2018-01-31
Partial transitive dependency import support

Kindrat:
Added property to allow partial import of transitive dependencies without breaking import execution. In my case it helped with importing projects with some missing dependencies - e.g. babel 0.8. All skipped entries are logged as errors so you could add them lately. By default, the behaviour of pivy-importer remains unchanged - lenient import should be enabled explicitly


Kindrat:
Any comments?

PR #196, opened 2018-03-10
Add the framework for package-specific settings.

zvezdan:
Add the interface and the default implementation for package specific
settings. Use the framework calls in pip install and build wheel tasks.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
@@ -19,9 +19,13 @@ import com.linkedin.gradle.python.PythonExtension
 import com.linkedin.gradle.python.extension.PythonDetails
 import com.linkedin.gradle.python.extension.WheelExtension
 import com.linkedin.gradle.python.plugin.PythonHelpers
+import com.linkedin.gradle.python.util.DefaultEnvironmentMerger
+import com.linkedin.gradle.python.util.DefaultPackageSettings
 import com.linkedin.gradle.python.util.DependencyOrder
+import com.linkedin.gradle.python.util.EnvironmentMerger
 import com.linkedin.gradle.python.util.ExtensionUtils
 import com.linkedin.gradle.python.util.PackageInfo
+import com.linkedin.gradle.python.util.PackageSettings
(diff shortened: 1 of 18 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
```diff
@@ -50,17 +58,24 @@ class BuildWheelsTask extends DefaultTask implements SupportsWheelCache {
...
 
+    @InputFiles
+    FileCollection installFileCollection
+
+    @Input
+    @Optional
+    Map<String, String> environment
+
+    @Input
+    @Optional
```
kopy07: Why is it optional? It has a default, would it be required?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/DefaultEnvironmentMerger.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Default implementation of EnvironmentMerger.
+ *
+ * Just updates maps directly.
+ */
+public class DefaultEnvironmentMerger implements EnvironmentMerger {
+    @Override
+    public Map<String, String> mergeEnvironment(Map<String, String> target, Map<String, String> source) {
+        if (source != null) {
```
kopy07: I think you probably want to make a new map, and insert all of the non-null values into it. This will change the origional I feel will cause problems down the road.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/PackageSettings.java
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Interface for package specific settings.
+ *
+ * The settings needed for a specific package to build properly
+ * or to customize its build can be obtained from the API provided
+ * by this interface.
+ *
+ * @param <T> Type of the object that represents package information.
+ */
+public interface PackageSettings<T> {
```
kopy07: Do we want to bound T to be something?

zvezdan: The intent is to have a generic type for future development. No need to change T in this interface when we change the package information type.

PR #197, opened 2018-03-12
Adding Wheel Filtering Support

ethankhall:
Sometimes, wheels should be skipped from caching. This is because we
don't want to re-use dependencies that can change, as they won't be
re-built afterwards.


sholsapp:
Can you explain more why do we need this change? How do I know that something shouldn't be cached?

ethankhall:
With #196 we don't need this. This was intended to make it so that changing (SNAPSHOT, DEV, etc) packages don't get stored into a cache, and not rebuilt. But with #196, @zvezdan make it where we can check if the dependency should be built instead of using a cached value, so I closed the PR since we don't need it.

PR #200, opened 2018-03-15
Upgrade the unit tests to use new fromPath method.

zvezdan:
Replace all usage of deprecated fromPath method with the new one.
Add one test that confirms that deprecated method still works until we
drop it.

### DIFF
--- a/pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/util/PackageInfoTest.groovy
+++ b/pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/util/PackageInfoTest.groovy
@@ -22,7 +22,7 @@ class PackageInfoTest extends Specification {
 
     def 'can parse boring sdist'() {
         when:
-        def packageInfo = PackageInfo.fromPath('foo-1.0.0.tar.gz')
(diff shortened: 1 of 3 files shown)


pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/util/PackageInfoTest.groovy
```diff
@@ -22,56 +22,65 @@ class PackageInfoTest extends Specification {
...
+        def packageInfo = PackageInfo.fromPath(new File('Z:\\pygradle\\build\\ivy-repo\\pypi\\setuptools\\19.1.1\\setuptools-19.1.1'))
         then:
         assert packageInfo.name == 'setuptools'
         assert packageInfo.version == '19.1.1'
     }
 
     def 'can not parse a sdist that has an unknown extension'() {
         when:
-        PackageInfo.fromPath('foo-1.0.0.xxx')
+        PackageInfo.fromPath(new File('foo-1.0.0.xxx'))
```
kopy07: What do you think about replacing the `new File` with `makePackage` that returns a file, should in the future we want to make it actually create a file on disk?

Issue #205, opened 2018-03-21
java.io.NotSerializableException: com.linkedin.gradle.python.util.OperatingSystem$Linux


chrisbrake:
When I was trying to build some custom tasks around building virtual environments I discovered that if I declared outputs, the task would fail.

I've created a branch for reference, and reproduced the issue in the [example-project](https://github.com/chrisbrake/pygradle/tree/master/examples/example-project).

Here's a sample of what the error looks like
```
$ gradle installPythonRequirementsInAnotherVenv

FAILURE: Build failed with an exception.

* What went wrong:
Unable to store input properties for task ':installPythonRequirementsInAnotherVenv'. Property 'pythonDetails' with value 'com.linkedin.gradle.python.extension.PythonDetails@601bcc16' cannot be serialized.
> java.io.NotSerializableException: com.linkedin.gradle.python.util.OperatingSystem$Linux

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 0s
6 actionable tasks: 2 executed, 4 up-to-date
```
 To suppress the error, just comment out [this line](https://github.com/chrisbrake/pygradle/blob/20747d4dc94ddb65e6fd4d479f10c502c85a7e02/examples/example-project/build.gradle#L48).

From what I can see OperatingSystem is never serializable, but Gradle only tries to serialize it when outputs are declared for the task.

chrisbrake:
I was playing around with this, this evening.  

[...]

PR #206, opened 2018-03-22
Fixed Java Serialization issue

ethankhall:
In #205 we found a regression. This change fixes the issue, and adds a
test so we won't run into it again.

### DIFF
--- a/pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PythonSerializableTest.groovy
+++ b/pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PythonSerializableTest.groovy
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2016 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


chrisbrake:
Tested and confirmed this does resolve the issue I reported in #205

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
```diff
@@ -97,13 +107,7 @@ class BuildWheelsTask extends DefaultTask implements SupportsWheelCache, Support
     /**
      * Will return true when the package should be excluded from being installed.
```
sixninetynine: is this comment still accurate?

zvezdan: Yes. It still documents what the intended usage of this filter is. It's still a correct documentation for a public field.

pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PythonSerializableTest.groovy
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright 2016 LinkedIn Corp.
```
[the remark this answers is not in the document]
zvezdan: License checker is rather blunt.
It wants all the files to have the same year.
We didn't spend too much time finding a better one. :-(

PR #211, opened 2018-03-26
Fix regex to allow wheel dependencies. Fixes #210

### LINKED ISSUE #210: Latest version of PyGradle breaks Wheels dependencies
With PyGradle 0.6.21, I used to be able to specify a wheel dependency in my build.gradle file like this: 
```
dependencies {
    python 'mygroup:my_python_module:0.0.8'
}
```
Now, when I try to upgrade to PyGradle 0.7.7, I get this error:
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':installPythonRequirements'.
> Cannot calculate Python package extension from C:\Users\adagenais\.gradle\caches\modules-2\files-2.1\mygroup\my_python_module\0.0.8\c18e1f71e7ee593edfbbeea335e15328a7f620d\my_python_module-0.0.8-py2-none-any.whl using regular expression /\.tar\.gz|\.zip|\.tar|\.tar\.bz2|\.tgz/.
```

ameliedagenais:
We used to be able to specify Python dependencies on wheels packages in the build.gradle file using the `python` configuration, but this functionality is no longer working. Issue #210 has all the details. 

Unfortunately, it seems like wheel dependencies were never officially supported, but happened to work because the regex wasn't used properly. This change fixed applying the regex, but unfortunately broke wheel dependency support as a result: https://github.com/linkedin/pygradle/commit/0a6b7b6aa96f56e82abf5db635b77c55c691e3cb#diff-f3ee7d4e9fd88d136b10b3a5262ce74fR78

Since wheel packages are a standard in the Python community and PyGradle can handle it, I would like to update the regex to allow it.

### DIFF
--- a/pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/util/PackageInfoTest.groovy
+++ b/pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/util/PackageInfoTest.groovy
@@ -75,6 +75,14 @@ class PackageInfoTest extends Specification {
         thrown(GradleException)
     }
 
+    def 'can parse a wheel distribution'() {
+        when:
+        def packageInfo = packageInGradleCache('foo-1.0.0-py2-none-any.whl')
+        then:
+        assert packageInfo.name == 'foo'
+        assert packageInfo.version == '1.0.0'
+    }
+
     // TODO: Remove this test completely when we drop the deprecated method.
     def 'still supports deprecated fromPath method'() {
         when:
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/PackageInfo.groovy
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/PackageInfo.groovy
@@ -15,6 +15,7 @@
  */
 package com.linkedin.gradle.python.util
 
+import com.linkedin.gradle.python.wheel.PythonWheelDetails
 import org.apache.commons.io.FilenameUtils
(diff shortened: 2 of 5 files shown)


pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/util/PackageInfoTest.groovy
```diff
@@ -75,6 +75,14 @@ class PackageInfoTest extends Specification {
         thrown(GradleException)
     }
 
+    def 'can parse a wheel distribution'() {
+        when:
+        def packageInfo = packageInGradleCache('foo-1.0.0-py2-none-any.whl')
+        then:
+        assert packageInfo.name == 'foo'
+        assert packageInfo.version == '1.0.0-py2-none-any'
```
zvezdan: The version is not extracted correctly.
It should be 1.0.0 only.
This would break later when we try to install `foo==1.0.0-py2-none-any` which wouldn't exist. Pip would require` foo==1.0.0` there.
If we want to enable wheels to be handled through this same interface, we'd have to change the approach here and use a different matching.
This was clearly intended for `sdist` only.
There's a code in pygradle that parses wheels and finds the name, version, ... correctly off them.
Perhaps you could look into using that API.

ameliedagenais: Agreed. I updated the pr accordingly.

PR #215, opened 2018-03-30
Fix the regex for wheel file name format.

zvezdan:
This fixes the regex to disallow empty parts.
It also takes care of converting correctly "_SNAPSHOT" back into
"-SNAPSHOT" (or similar semver.org pre-release formats).


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/wheel/PythonWheelDetails.java
```diff
@@ -35,10 +38,17 @@
     public PythonWheelDetails(File file, Matcher matcher) {
         this.file = file;
         this.dist = matcher.group("dist");
-        this.version = matcher.group("version");
+        String version = matcher.group("version");
```
kopy07: lets use a name other than version, shadowing like this confuses me :-)

PR #217, opened 2018-04-04
Fix handling of project name and directory.

zvezdan:
If a meta-package matches a dependency package name, it may get
`--editable` during the install of the dependency and fail.

This change ensures that we're processing the project directory by
confirming that all of the following are true: version is not set,
package file is a directory, package directory matches the project
directory.  The last check is for packages that have projects named
containing-dir/backend and package name ends up being just backend.

The order of checks is intentional to cut checks as early as possible.
Hence the version check first, since other packages will have it set.


sixninetynine:
Does this mean that your subproject directory and project name must strictly match?

kopy07:
@sixninetynine it used to, now it shouldn't.

PR #237, opened 2018-07-10
Refactored IntegTest to become unit test.

ethankhall:
PackageSettingsIntegrationTest was flakey and really just wanted to
ensure that specific arguments were passed to pip. So now of running a
build, we will validate that the arguments would have been passed to
pip. This makes the tests fast and not rely on the env. It also means
that we don't need to do any strange regex's and rely on side affects of
the build processes to validate things.

The previous test was asserting that Gradle was passing args to the
commandline, and that's the same as asserting that our code was calling
Gradle with the correct parameters.

This change splits the integ test into two tests, because it was testing
two different things. `pip install` and `pip wheel`, so now there is one
test for each.

To do this, two actions were extracted. PipInstall and PipWheel. They
both use the same base class so we can re-use as much code as possible.


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PipInstallTask.groovy
```diff
@@ -129,27 +125,39 @@ class PipInstallTask extends DefaultTask implements FailureReasonProvider, Suppo
...
+                progressLogger.progress("Installing ${packageInfo.toShortHand()} (${++counter} of ${installableFiles.size()})")
+
+                if (packageExcludeFilter != null && packageExcludeFilter.isSatisfiedBy(packageInfo)) {
+                    if (PythonHelpers.isPlainOrVerbose(project)) {
+                        logger.lifecycle("Skipping {} - Excluded", packageInfo.toShortHand())
+                    }
+                } else {
+                    try {
+                        pipInstallAction.installPackage(packageInfo, args)
+                    } catch (PipExecutionException pee) {
```
zvezdan: Let's call it just `e`. :-)

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/extension/PythonDetailsTestDouble.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package com.linkedin.gradle.python.extension;
+
+import org.gradle.api.Project;
+
+import java.io.File;
+
+public class PythonDetailsTestDouble extends PythonDetails {
+
+    private PythonVersion version = new PythonVersion("3.4");
```
zvezdan: I know this is just a test, but since we never used 3.4, can we please use something higher, 3.6 or 3.5 at least, even as an example in the test.

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/tasks/action/PipActionHelpers.groovy
```diff
@@ -0,0 +1,116 @@
...
+            this.envOverride = envOverride
+        }
+
+        @Override
+        Map<String, String> getEnvironment(PackageInfo packageInfo) {
+            envOverride.getOrDefault(packageInfo.name, [:])
+        }
+    }
+
+    static class BuildOptionsPackageSettings extends DefaultPackageSettings {
```
zvezdan: For consistency, this should also be a singular "Option" (or other names need to have plural), and should have Override in the name as other classes do.

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/tasks/action/PipInstallActionTest.groovy
```diff
@@ -0,0 +1,138 @@
...
+        def pipInstallAction = createPipInstallAction(settings, execSpec)
+
+        when:
+        pipInstallAction.installPackage(packageInGradleCache("setuptools-1.0.0.tar.gz"), [])
+
+        then:
+        1 * execSpec.commandLine(_) >> { List<List<String>> args ->
+            println args
+            def it = args[0]
+            def idx = it.indexOf('--dummy-global-option')
```
zvezdan: Let's change the order here and search for `--global-option` with the confirmation that `--dummy-global-option` follows it.

Also, this being Groovy, isn't there a more elegant way to compare slices of the list?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/exception/PipInstallException.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package com.linkedin.gradle.python.exception;
+
+import com.linkedin.gradle.python.util.PackageInfo;
+import org.gradle.api.GradleException;
+
+import java.util.List;
+
+public class PipInstallException extends GradleException {
```
zvezdan: This class is not used anywhere and its only method seems to be a duplicate of PipExecutionException equivalent method. Perhaps we need to get rid of it?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/action/pip/PipWheelAction.java
```diff
@@ -0,0 +1,138 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public class PipWheelAction extends AbstractPipAction {
```
zvezdan: This class is not used anywhere.
The `PipInstallAction` was refactored from` PipInstallTask` and used there, AFAICS.
OTOH, this class seems to be refactored from `BuildWheelsTask`, but that class was not changed in this PR and this class is not used in it. Omission?

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/tasks/action/PipInstallActionTest.groovy
ethankhall: Possible... But then I'de have to compare the whole thing I think.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/action/pip/PipWheelAction.java
ethankhall: Yep. I missed it. Will add w/ next update.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/VirtualEnvironment.java
```diff
@@ -51,16 +51,20 @@ public static String getPythonApplicationDirectory() {
         return OperatingSystem.current().isWindows() ? "Scripts" : "bin";
     }
 
-    public File getExecutable(String path) {
-        return validateFileExists(prefixBuilder().resolve(OperatingSystem.current().getExecutableName(path)));
+    public File getExecutable(String execName) {
```
zvezdan: The type is the same here (String), but the semantics is not (name vs. path).
Are we sure that existing usage of this method and the `findExecutable` below will be OK with this change in semantics?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
```diff
@@ -18,40 +18,29 @@ package com.linkedin.gradle.python.tasks
...
+import com.linkedin.gradle.python.tasks.exec.ProjectExternalExec
 import com.linkedin.gradle.python.tasks.supports.SupportsPackageInfoSettings
 import com.linkedin.gradle.python.tasks.supports.SupportsWheelCache
-import com.linkedin.gradle.python.util.DefaultEnvironmentMerger
-import com.linkedin.gradle.python.util.DependencyOrder
-import com.linkedin.gradle.python.util.EnvironmentMerger
-import com.linkedin.gradle.python.util.ExtensionUtils
-import com.linkedin.gradle.python.util.PackageInfo
-import com.linkedin.gradle.python.util.PackageSettings
+import com.linkedin.gradle.python.util.*
```
zvezdan: No star imports please.
Revert this part to use full imports of classes from util as before.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
```diff
@@ -147,105 +137,22 @@ class BuildWheelsTask extends DefaultTask implements SupportsWheelCache, Support
...
+            pythonDetails, wheelCache, environmentMerger, wheelExtension)
+
         def taskTimer = new TaskTimer()
 
         int counter = 0
         def numberOfInstallables = installables.size()
         installables.each { File installable ->
-            def pyVersion = pythonDetails.getPythonVersion().pythonMajorMinor
             def packageInfo = PackageInfo.fromPath(installable)
             def shortHand = packageInfo.version ? "${ packageInfo.name }-${ packageInfo.version }" : packageInfo.name
```
zvezdan: The `PipInstallTask` uses the new `PackageInfo.toShortHand()` method for this.
Can we reuse the same here instead of recalculating?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
```diff
@@ -147,105 +137,22 @@ class BuildWheelsTask extends DefaultTask implements SupportsWheelCache, Support
...
-            }
-
-            if (installResult.exitValue != 0) {
-                LOGGER.error("Error installing package using `{}`", commandLine)
-                LOGGER.error(stream.toString().trim())
-                throw new GradleException("Failed to build wheel for ${ shortHand }. Please see above output for reason, or re-run your build using ``--info`` for additional logging.")
-            } else {
-                LOGGER.info(stream.toString().trim())
-            }
+            wheelAction.buildWheel(packageInfo, args)
```
zvezdan: The `PipInstallTask` has a check for `packageExcludeFilter` here in case the dependency is conditional for a certain Python version and needs to be excluded.

We do not do that check neither here, nor in the `wheelAction.buildWheel()` method.
That needs to be fixed to have the code that is equivalent to the old one.

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/tasks/action/PipActionHelpers.groovy
```diff
@@ -0,0 +1,116 @@
...
+        private final List<String> override
+
+        RequiresRebuildOverridePackageSettings(TemporaryFolder temporaryFolder, List<String> override) {
+            super(temporaryFolder.root)
+            this.override = override
+        }
+
+        @Override
+        boolean requiresSourceBuild(PackageInfo packageInfo) {
+            return override.contains(packageInfo.name)
```
zvezdan: To match the deleted integration tests and semantic intent, this methods and other methods in this test class should call super when not matching the package name. Otherwise, any future change in the default class will **not** match these tests. I believe this should be `return (override.contains(packageInfo.name)) ? true : super.requiresSourceBuild(packageInfo)`. Similarly for others.

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/tasks/action/PipActionHelpers.groovy
```diff
@@ -0,0 +1,116 @@
...
+            this.envOverride = envOverride
+        }
+
+        @Override
+        Map<String, String> getEnvironment(PackageInfo packageInfo) {
+            envOverride.getOrDefault(packageInfo.name, [:])
+        }
+    }
+
+    static class BuildOptionPackageSettings extends DefaultPackageSettings {
```
zvezdan: `BuildOptionOverridePackageSetting` to be consistent with other names?

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/tasks/action/PipWheelActionTest.groovy
```diff
@@ -0,0 +1,143 @@
...
+
+import static com.linkedin.gradle.python.tasks.action.pip.PipActionHelpers.packageInGradleCache
+
+
+class PipWheelActionTest extends Specification {
+
+    @Rule
+    TemporaryFolder temporaryFolder
+
+    def "wheel install uses environment"() {
```
zvezdan: Please, remove the word "install" here and in other tests below.
The command given to pip in these executions is "wheel" and that word is enough to describe that this command does the given described action. The "install" was needed above in the "pip install" tests.

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/tasks/action/PipWheelActionTest.groovy
```diff
@@ -0,0 +1,143 @@
...
+        action.buildWheel(packageInGradleCache("pyflakes-1.0.0.tar.gz"), [])
+
+        then:
+        1 * execSpec.commandLine(_) >> { List<List<String>> args ->
+            println args
+            def it = args[0]
+            def idx = it.indexOf('--no-deps')
+            assert idx != -1
+            assert it[2] == 'wheel'
+            assert !it.any { entry -> entry == '--ignore-installed' }
```
zvezdan: Sure, but this is not confirming that the wheel was actually required to be rebuilt/reinstalled.
This test prompted me to put the comments above about lack of integration test quality with these tests.

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/tasks/action/PipInstallActionTest.groovy
```diff
@@ -0,0 +1,140 @@
...
+import com.linkedin.gradle.python.wheel.EmptyWheelCache
+import org.gradle.process.ExecSpec
+import org.gradle.testfixtures.ProjectBuilder
+import org.junit.Rule
+import org.junit.rules.TemporaryFolder
+import spock.lang.Specification
+
+import static com.linkedin.gradle.python.tasks.action.pip.PipActionHelpers.packageInGradleCache
+
+class PipInstallActionTest extends Specification {
```
zvezdan: All of these tests just confirm that the arguments will be correctly passed if the action gets actually called.
They are however not a replacement for integration tests because they cannot actually prove that the action was ever called in real integration.

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/tasks/action/PipWheelActionTest.groovy
```diff
@@ -0,0 +1,143 @@
...
+import org.gradle.process.ExecSpec
+import org.gradle.testfixtures.ProjectBuilder
+import org.junit.Rule
+import org.junit.rules.TemporaryFolder
+import spock.lang.Specification
+
+import static com.linkedin.gradle.python.tasks.action.pip.PipActionHelpers.packageInGradleCache
+
+
+class PipWheelActionTest extends Specification {
```
zvezdan: See the comment above about why these cannot replace the integration tests.

pygradle-plugin/src/integTest/groovy/com/linkedin/gradle/python/plugin/PackageSettingsIntegrationTest.groovy
```diff
@@ -1,594 +0,0 @@
-/*
```
zvezdan: See the comment below on the new tests on why I think they do not completely replace integration tests and by their nature cannot replace them fully.

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/tasks/action/PipWheelActionTest.groovy
ethankhall: Fixed. Made the `WheelCache` a mock, and asserting there are no interactions with it.

pygradle-plugin/src/test/groovy/com/linkedin/gradle/python/tasks/action/PipActionHelpers.groovy
ethankhall: I don't think that's needed. We are trying to force the behavior for the tests, so we don't really care if these tests don't do what the super class does.

If the behavior of the default class changes, then the tests should be updated there. These tests should be isolate from changes to DefaultPackageSettings.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/extension/VirtualEnvironment.java
ethankhall: It was never really a path, it was always something like "pip" or "pex". So I think the name change is better. Pip and Pex were found during tests/integTests so I think it will be OK.

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
ethankhall: Done

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/BuildWheelsTask.groovy
ethankhall: Fixed. I missed that.

PR #245, opened 2018-08-28
Update mock dependency version

Ricky-N:
LinkedIn's [public Artifactory](https://linkedin.jfrog.io/linkedin/webapp/#/artifacts/browse/tree/General/pypi-external/pypi/mock) no longer contains the `1.0.1` version of this mock dependency. This upgrades it so the getting started can keep flowing smoothly


zvezdan:
@Ricky-N Thank you for the update in the docs!

PR #247, opened 2018-09-21
Upgrade to pex-1.4.6.

zvezdan:
This finally allows us to move to newer setuptools and wheel that are
able to parse the new syntax for conditional dependencies in
install_requires instead of extras_require in setup of Python packages.
Using the opportunity to upgrade virtualenv, so that its vended versions
of setuptools, wheel, and pip better match the new ones. Finally,
upgrading to the latest pip too.


sixninetynine:
Let's try and be sure we don't get bit by these issues https://github.com/pantsbuild/pex/issues/550

PR #269, opened 2018-12-07
Added lenient decorator for Pypi cache to decouple args specific behaviour

Kindrat:
The goal is to wrap all arg-specific processing (prerelease, extras, lenient, etc) with decorators to simplify common services and provide more flexibility for future extension. This is first part covering PypiApiCache only


Kindrat:
Would appreciate if anyone could look at this and say if it can be merged to upstream

PR #275, opened 2019-02-05
Extend package settings API to support `isCustomized` checking.

mhbkb:
The new API can be useful for checking if the binary artifact
of the package needs to rebuild if it doesn't exist in local
project cache.

Note that if the project is customized, it won't read and write
from global cache.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/PackageSettings.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/PackageSettings.java
@@ -94,11 +94,19 @@ public interface PackageSettings<T> {
      * Determines if the package requires a build from source.
      * <p>
      * Even if the binary artifact of the package is available,
-     * the package may need a rebuild because of custom build
-     * options or environment.
+     * the package may need a rebuild because of editable current
+     * project or snapshots.
      *
      * @param t package information object
      * @return true when the package needs a rebuild from source
      */
     boolean requiresSourceBuild(T t);
+
+    /**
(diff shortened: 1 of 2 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/PackageSettings.java
```diff
@@ -91,14 +91,22 @@
...
 
     /**
-     * Determines if the package requires a build from source.
+     * Determine if the package always requires a build from source.
      * <p>
-     * Even if the binary artifact of the package is available,
-     * the package may need a rebuild because of custom build
-     * options or environment.
+     * Even if the binary artifact of the package is available in
+     * both global and local cache, the package may need a rebuild
```
zvezdan: No need to explain the caching here or number of layers (global/local). Let's keep the previous description and just change the part after "because ..."

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/PackageSettings.java
```diff
@@ -91,14 +91,22 @@
     List<String> getSupportedLanguageVersions(T t);
 
     /**
-     * Determines if the package requires a build from source.
```
zvezdan: Let's keep the descriptive word here "Determines" because that's Javadoc style -- this function "determines ...".
In some other languages the imperative style is preferred as if giving a command to the function: "Determine ..."
Similarly below, using "Checks ... " is probably more fitting to this style.

PR #278, opened 2019-02-08
Make a new wheel utility class.

mhbkb:
Add makeWheelFromSdist function to the new wheel utility class.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/WheelUtils.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/WheelUtils.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2016 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.linkedin.gradle.python.util;
+
+import com.linkedin.gradle.python.extension.PythonDetails;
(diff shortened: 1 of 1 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/WheelUtils.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.File;
+import java.io.IOException;
+import org.apache.commons.io.FileUtils;
+import org.gradle.api.Project;
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
+import org.gradle.process.ExecResult;
+
+
+public class WheelUtils {
```
zvezdan: The class is a utility class in the sense that it provides some functions (static methods in absence of better in Java), but we don't need to call it `Utils`. Do we plan to add other functions here? If yes, perhaps a full name `WheelUtilities` would be nicer. If not, perhaps just its current purpose should do. Something like `WheelMaker` perhaps?

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/util/WheelUtils.java
```diff
@@ -0,0 +1,80 @@
...
+     * @param pythonDetails details on the python to build the wheel
+     * @param targetDir target directory to store the built wheel
+     */
+    public static void makeWheelFromSdist(Project project, PackageInfo packageInfo, PythonDetails pythonDetails, File targetDir) {
+        if (packageInfo.getPackageFile().getName().endsWith(".whl")) {
+            return;
+        }
+
+        ByteArrayOutputStream stream = new ByteArrayOutputStream();
+        ExecResult results = project.exec(exec -> {
```
zvezdan: I wonder if we could get a pip action class (or its part) to do this instead of duplicating the work here?
Perhaps it's worth looking into refactoring, not just to avoid the duplication but to limit the number of places where we control such external executables and possible fixes in the future.

If you feel this will be temporary while we execute the incremental changes, that's fine.
In that case this class could be it.
We would need some sort of testability added to it in that case.
See examples of use of `ExternalExec` and `ProjectExternalExec`.

mhbkb:
After taking a good at existing code base, looks like we can re-use execCommand in AbstractPipAction.
Closing this pull request.

PR #279, opened 2019-02-12
Adapt pip install action to always try wheel first.

mhbkb:
Adapt pip install action to always try wheel first. If wheel
doesn't exist, try to make wheels. If failed, fall back to current
installation that tries to install from sdist.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/action/pip/PipInstallAction.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/action/pip/PipInstallAction.java
@@ -121,15 +121,17 @@ public class PipInstallAction extends AbstractPipAction {
         }
     }
 
-    private List<String> makeCommandLine(PackageInfo packageInfo, List<String> extraArgs) {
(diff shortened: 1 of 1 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/action/pip/PipInstallAction.java
```diff
@@ -121,7 +121,7 @@ void doPipOperation(PackageInfo packageInfo, List<String> extraArgs) {
         }
     }
 
-    private List<String> makeCommandLine(PackageInfo packageInfo, List<String> extraArgs) {
+    private List<String> makeCommandLine(PackageInfo packageInfo, List<String> extraArgs, boolean allowBuildingFromSdist) {
```
zvezdan: We don't need to change the signature of this method.
In Java, we can add another method with a different signature, but the same name, that would call this one and cause a different behavior. For example, by setting a private field that holds the state. The method with the shorter, default signature, can reset the state to the default at the end of its execution.
Alternatively, we could add another method with a different name to avoid a boolean argument that usually indicates the code smell.

PR #280, opened 2019-02-12
Adapt pip install task to invoke wheel building when not found.

mhbkb:
Add an extra allowBuildingFromSdist parameter in AbstractPipAction
execute function. It allows PipInstallTask to pass
allowBuildingFromSdist to PipInstallAction so that it can try to
install from wheel, and then fall back to install from sdist to
keep current behavior.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/action/pip/AbstractPipAction.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/action/pip/AbstractPipAction.java
@@ -65,7 +65,7 @@ abstract class AbstractPipAction {
         this.packageExcludeFilter = packageExcludeFilter;
     }
 
-    public void execute(PackageInfo packageInfo, List<String> extraArgs) {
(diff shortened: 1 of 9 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/action/pip/AbstractPipAction.java
```diff
@@ -65,15 +65,15 @@
         this.packageExcludeFilter = packageExcludeFilter;
     }
 
-    public void execute(PackageInfo packageInfo, List<String> extraArgs) {
+    public void execute(PackageInfo packageInfo, List<String> extraArgs, boolean allowBuildingFromSdist) {
```
zvezdan: See also the comment on the other PR about using Java's ability to define two method signatures using the same name. Then you don't need to change the default calls and test calls. You can add new test calls of course as we can proceed with incremental changes.

mhbkb:
Closing this PR as https://github.com/linkedin/pygradle/pull/279 is cleaner to adapt wheel retrieving, installation and sdist fall backing.

PR #304, opened 2019-05-30
Update link to Gradle user guide

pioterj:


zvezdan:
Thanks for the fix. I'm going to re-run the failed Travis checks one more time.
Frankly, this change should not have caused any failures on Travis. It must be temporary flakiness.

PR #325, opened 2019-11-22
Added a null check for the htmlDir in PyCoverageTask

elautz:
Handling of an edge case: If the user configures the project
to run covergae but there is no coverage output (i.e. no
coverage to report), the htmlDir parsed from the output will
ultimately be null, and thus the call to project.file(htmlDir)
fails with IllegalArgumentException: path may not be null or
empty string. path='null'

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PyCoverageTask.groovy
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PyCoverageTask.groovy
@@ -73,8 +73,10 @@ class PyCoverageTask extends PyTestTask {
         String htmlDir = streamProcessor.htmlDir
         String coverage = streamProcessor.coverage
 
-
-        FileUtils.copyDirectoryToDirectory(project.file(htmlDir), coverageOutputDir)
+        // If there is no coverage to report, then the htmlDir value will be empty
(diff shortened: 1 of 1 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/PyCoverageTask.groovy
```diff
@@ -73,8 +73,10 @@ class PyCoverageTask extends PyTestTask {
         String htmlDir = streamProcessor.htmlDir
         String coverage = streamProcessor.coverage
 
-
-        FileUtils.copyDirectoryToDirectory(project.file(htmlDir), coverageOutputDir)
+        // If there is no coverage to report, then the htmlDir value will be empty
+        if ((htmlDir != null) && !htmlDir.isEmpty()) {
```
zvezdan: Do we need to worry about empty directory?
It seems that it would be OK to have an empty directory copied into the output directory.
Perhaps I'm missing something in that copying behavior, but a check for `null` could be enough, couldn't it?

elautz: You're right. I just confirmed that the fix works just fine with only a null check. I updated the commit accordingly.

PR #333, opened 2019-12-08
Added infrastructure to safely bump flake8 version

elautz:
Added logic in the Flake8Task to rerun the execution with ignoring
a set of rules if the first execution fails. It will rerun if the
IGNORE_RULES set it non-empty. This set of rules will contain
any rules/checks added between flake8 versions. For example, when
we bump to flake8 3.8.0, we should update the IGNORE_RULES value
to hold all rules and checks added between flake8 3.6.0 and 3.8.0.

If the second run, when ignoring newly added rules, succeeds, we
now emit a warning to the user stating which rules are being
ignored, and which of these ignored rules failed for them. Added a
method called overrideIgnoreRules() which is used for integration
testings and will be used internally by LI. Integration tests
were added to verify this new functionality.

### DIFF
--- a/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/Flake8Task.java
+++ b/pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/Flake8Task.java
@@ -17,6 +17,8 @@ package com.linkedin.gradle.python.tasks;
 
 import com.linkedin.gradle.python.PythonExtension;
 import com.linkedin.gradle.python.util.ExtensionUtils;
+import java.util.HashSet;
+import java.util.Set;
 import org.apache.commons.io.FileUtils;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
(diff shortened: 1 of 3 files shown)


pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/Flake8Task.java
```diff
@@ -32,8 +34,37 @@
 public class Flake8Task extends AbstractPythonMainSourceDefaultTask {
 
     private static final Logger log = Logging.getLogger(Flake8Task.class);
+    // Track whether the current run is excluding the new rules
+    private Boolean ignoringNewRules = false;
+    private String firstRunOutput = null;
+
+    /* Set of Flake8 rules that were added in version bumps. This set of rules, and this comment should be updated
```
zvezdan: Let's use the commenting style for long (paragraph) comments that's same as in the rest of the code:
```
/*
 * Set of flake8 rules ... version updates.
 * ...
 */

File: pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/Flake8Task.java [hunk 2]
```diff
@@ -32,8 +34,37 @@
...
+       every time the flake8 version is bumped.
```
zvezdan: s/bumped/updated/

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/Flake8Task.java
```diff
@@ -32,8 +34,37 @@
...
+       Rules added between versions:
+     */
+    private static final Set<String> IGNORE_RULES = new HashSet<>(Arrays.asList(
+        // Empty for now until we bump to flake8 3.8.0
+    ));
+
+    private Set<String> ignoreRules = IGNORE_RULES;
+
+    private static final String IGNORED_RULES_MSG = "######################### WARNING ##########################\n"
+        + "The flake8 version has recently been bumped, which add the following new rules:\n"
```
zvezdan: ... has been recently updated, which added ...

pygradle-plugin/src/main/groovy/com/linkedin/gradle/python/tasks/Flake8Task.java
```diff
@@ -32,8 +34,37 @@
...
+        + "%s\n"  // This will be replaced with the set of ignored rules
+        + "Your project is failing for one or more of these rules. Please address them, as they will be enforced soon.\n"
+        + "%s############################################################\n";
+
+    // Provide the ability to override the ignored rules for testing purposed, but mainly so the LI side can maintain
```
zvezdan: s/purposed/purposes/

others may need to do similar things besides us, so I'd change the rest of the comment to: "so that users can enforce the use of a different version of flake8 in their plugins."

PR #341, opened 2020-03-05
Ivy and PyPi module, file names compatibility

BrainLogic:
Fix for the following issue:
for PiPy modules that have non-equal file and module names, ivy metafiles will be generated incorrectly.
Because when Importer generates dependencies part of ivy file, it takes module name from name field of PyPi meta but when Importer creates directories and artifact file itself, directory name and file name are created from filename field of PyPi meta, and these two configuration aspects can be different. It means that ivy cannot find files for specified dependencies.

for instance, I was trying to add `mypy` library https://pypi.org/pypi/mypy/json
Importer created dependencies like this:
``` 
<dependencies defaultconfmapping="*->default">
    <dependency org="pypi" name="typed-ast" rev="1.4.0" conf="default" />
    <dependency org="pypi" name="typing-extensions" rev="3.7.4" conf="default" />
    <dependency org="pypi" name="mypy-extensions" rev="0.4.3" conf="default" />
</dependencies>
```
but directories and file names have been created with underscore `mypy_extensions`.

In the fix, module name, directory name and file name are created from the same value of the `name` field of PyPi metafile (ex. https://pypi.org/pypi/mypy/json).

### DIFF
--- a/pivy-importer/src/main/groovy/com/linkedin/python/importer/deps/DependencyDownloader.groovy
+++ b/pivy-importer/src/main/groovy/com/linkedin/python/importer/deps/DependencyDownloader.groovy
@@ -18,7 +18,6 @@ package com.linkedin.python.importer.deps
 import com.linkedin.python.importer.pypi.PypiApiCache
 import com.linkedin.python.importer.util.ProxyDetector
 import groovy.util.logging.Slf4j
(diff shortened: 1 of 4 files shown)


BrainLogic:
@zvezdan please help to review and merge this PR

PR #354, opened 2021-12-17
working example project of pygradle

pash0002:
…pi repository.


pash0002:
Folks, Seems that current examples not working. So I've added the working example of the **`(py)gradle`** project.

PR #356, opened 2022-04-25
Gradle 7 Compatibility

Cdevon2:
- Add missing annotations to make pygradle plugin tasks compatible with Gradle 7
- Fix issues with extra characters when writing wheel-api.py resource to file
- Update TeeOutputstream constructor to refer to static System.out instead of using self-referential assignment


Cdevon2:
The failing test is due to updates in the underlying docker image.  The openjdk:8-jdk default image uses Debian Bullseye as the base, while the default image in the previous release used Debian Buster.  Bullseye does not provide a default python executable.

Oloremo:
Any updates on this?

